PR #5, opened 2016-06-10
Fix issues with numbered lists in README when rendered.

Fryguy:
Before, the numbered lists would restart their numbering in the middle.
After, the numbered lists continue to increase the number.


gregnavis:
Great catch! Merging. Thank you for contributing.

PR #14, opened 2018-01-12
Extraneous index check can skip fulltext indexes

tcopeland:
ActiveRecord::ConnectionAdapters::IndexDefinition#columns returns a
String in the case of PostgreSQL full-text indexes, e.g.:

"to_tsvector('english'::regconfig, (first_name)::text)"

So those can be skipped for this analyzer.  Otherwise we get
ArgumentError: wrong number of arguments (given 0, expected 1+)
exceptions from ExtraneousIndex#prefix?.


tcopeland:
The build failed, but if I'm reading the Travis logs right that's because of a missing gem issue, not because of this code change.  Although, I didn't do any tests for this... shame on me!

Issue #21, opened 2019-01-03
missing_presence_validation is not needed for belongs_to in Rails 5


duleorlovic:
In [Rails 5](https://github.com/rails/rails/issues/18233) `belongs_to` automatically adds presence validation.
The problem is that we check for [company_id](https://github.com/gregnavis/active_record_doctor/blob/master/lib/active_record_doctor/tasks/missing_presence_validation.rb#L34) but Rails add presence validation for `company`.

Would be enough to check for existence of validation without `_id` ?

duleorlovic:
I'm not able to create failing test for this. II do not know why those automatic validator do not show up, probably `Temping` is different than Rails `ApplicationRecord` in this case.
```
    Temping.create(:companies, temporary: false)
    Temping.create(:users, temporary: false) do
      with_columns do |t|
        t.references :company
        t.string :name
      end
    end
    User.class_eval do
      belongs_to :company
    end

 User.validators # => []
```

djezzzl:
@duleorlovic It should be showed up in `User.validators`, check this configuration:
`Rails.application.config.active_record.belongs_to_required_by_default = false`. And what is your Rails version? 

BTW, the check you want to have is implemented in my alternative: [database_consistency](https://github.com/djezzzl/database_consistency). Let me know if it solves your problem.

gregnavis:
Thank you for reporting, @duleorlovic!

Your test case didn't work because you need to specify `belongs_to :company, required: true` explicitly to ensure it's not overridden by defaults (as per @djezzzl comment). Also, you can put the block of `class_eval` right into the block passed to `Temping.create`.

The primary issue here is that in case of a `belongs_to` association validation can be defined on the association instead of the primary key. This affected two tests: `missing_presence_validation` and `missing_non_null_constraint`. Both would report incorrect results.

I prefer a more robust solution than your #22 and implemented it in https://github.com/gregnavis/active_record_doctor/commit/4e4e20914a00081878768910b9b1938af8f59c94 and https://github.com/gregnavis/active_record_doctor/commit/a4049b98fcf3f37925e2e520cb94203c52ecaa18. Instead of looking for the `_id` suffix we should reflect on the defined associations and match their names with foreign keys. Please see the linked commits for details.

Thank you for reporting the issue! I haven't released a new version yet due to issues with the CI server but feel free to pull from `master` and let me know whether these commits fixed the problem for you. If not, we'll look further.

gregnavis:
[...]

PR #22, opened 2019-01-03
Check if presence validation if on column without _id #21

duleorlovic:


gregnavis:
Thank you for reporting and submitting a PR. As per [my explanation](https://github.com/gregnavis/active_record_doctor/issues/21#issuecomment-453120693), I prefer a more robust solution and hence my decision to close the PR. I wanted to fix that quickly and hope you don't mind that. I'll of course give you credit in `CHANGLOG.md`.

PR #35, opened 2020-01-13
Fix missing_non_null_constraint on null: true column with association

chubchenko:
```bash
1) Failure:
ActiveRecordDoctor::Tasks::MissingNonNullConstraintTest#test_presence_true_with_null_true [active_record_doctor/test/active_record_doctor/tasks/missing_non_null_constraint_test.rb:114]:

Expected: {}
Actual: {"users"=>["name"]}
```

### DIFF
--- a/lib/active_record_doctor/tasks/missing_non_null_constraint.rb
+++ b/lib/active_record_doctor/tasks/missing_non_null_constraint.rb
@@ -38,7 +38,7 @@ module ActiveRecordDoctor
 
         model.validators.any? do |validator|
           validator.is_a?(ActiveRecord::Validations::PresenceValidator) &&
-            (validator.attributes && allowed_attributes).present? &&
+            (validator.attributes & allowed_attributes).present? &&
             !validator.options[:allow_nil] &&
             !validator.options[:if] &&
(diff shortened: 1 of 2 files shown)


chubchenko:
Currently, I have 1 failed build with ruby `2.2.10` and rails `5.2.x`. It's related to the latest release of rails where was [added](https://github.com/rails/rails/blob/v5.2.4.1/actionpack/lib/action_dispatch/request/session.rb#L96) a safe navigation operator which does not support by ruby < 2.3.x.
It's already [fixed](https://github.com/rails/rails/blob/5-2-stable/actionpack/lib/action_dispatch/request/session.rb#L96) in the 5-2-stable branch but the release has not yet been released.

gregnavis:
@chubchenko thank you for going through the trouble of debugging and fixing the problem.

There was already a similar test case so I took the liberty of editing your change slightly which resulted in this commit: https://github.com/gregnavis/active_record_doctor/commit/dae500a43a974ff5357115ffd4b40f27a5ea8739. Naturally, I retained all authorship information to give credit where it's due.

Thank you again for the contribution.

PR #37, opened 2020-01-30
Fix "undefined method `incorrect_boolean_presence_validation'"

eitoball:
This PR should fix `NoMethodError` when running `active_record_doctor:incorrect_boolean_presence_validation` task.


gregnavis:
@eitoball, many thanks for your contribution! I replaced the test you wrote with a test to ensure all tasks have appropriately named printers.

Thanks again!

PR #44, opened 2020-09-06
Add support of discarded_at to unindexed_deleted_at task

### LINKED ISSUE #42: Suggestion: paranoia is deprecteted / check for 'discarded_at' column as well
I would suggest modifying 'active_record_doctor:unindexed_deleted_at' to also check for the 'discarded_at' column because paranoia is more or less deprecated in favor of the 'discard' gem.  See the notice at the top of paranoia's readme [here](https://github.com/rubysherpas/paranoia).

dmitryzuev:
## Solution
Add both `deleted_at` and `discarded_at` to the recipe.

Closes #42


gregnavis:
Merged. @dmitryzuev, thank you for the contribution.

PR #45, opened 2020-09-11
Not all validators respond to attributes

maschwenk:
```ruby
class Blah < ActiveModel::Validator
  def validate
  end
end
```

Not all validator subclasses respond to `attributes`. In these cases I think it's just easiest to skip em. Happy to add tests if you agree with the change.

### DIFF
--- a/lib/active_record_doctor/tasks/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/tasks/missing_unique_indexes.rb
@@ -15,12 +15,12 @@ module ActiveRecordDoctor
             model.table_name,
             model.validators.select do |validator|
               table_name = model.table_name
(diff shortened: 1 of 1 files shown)


gregnavis:
@maschwenk, thank you for the PR. It'd be great if you could add a test case to the PR.

gregnavis:
Thanks for the contribution. I did some slight edits to your commit and commited it to master. You can see it in https://github.com/gregnavis/active_record_doctor/commit/5647b456ad217a55428a90c2dcf4d2f40d121724. Naturally, I retained the original authorship information.

Thanks again!

maschwenk:
@gregnavis appreciate it, sorry work week got in the way 😦

PR #49, opened 2020-10-01
Columns ordering in missing unique index validation should not matter

rhymes:
Right now if the uniqueness validator is `[:a, :b, :c]` but the unique index is `[:a, :c:, :b]` this checks fails, but the order is not strictly important. It can be regarding how databases behave in their query plan, for the Rails purposes is not important.

I need help with the tests, should I just add a new case?


gregnavis:
Merged in 75491c1. I took the liberty of adding a test case and providing a bit of a description in the commit message.

Thank you for the contribution.

rhymes:
@gregnavis thank you!

PR #52, opened 2021-02-12
Add compatibility with pg_search

kopylovvlad:
Hello!

I have noticed that detecting missing presence validations and detecting missing non-`NULL` constraints don't have compatibility with [pg_search](https://github.com/Casecommons/pg_search) gem. When I run rake-task I get error 

```
Caused by:
PG::UndefinedTable: ERROR:  relation "pg_search_documents" does not exist
LINE 5:                WHERE a.attrelid = '"pg_search_documents"'::r...
```

It is the easiest workaround to solve the problem.

What do you think about it? Maybe you have a better idea :)


gregnavis:
Thank you for the contribution, @kopylovvlad. We've fixed the problem in 8aa26ff1d6cfe182063b6e8ccd6ee9ad57bc4da1. It should be released as part of 1.7.2 this week.

PR #55, opened 2021-04-01
Use shorter name if index name is too long

duleorlovic:
To prevent exceptions like
```
StandardError: An error has occurred, all later migrations canceled:

Index name 'index_deleted_subscriber_invoice_status_transitions_on_subscriber_invoice_id' on table 'deleted_subscriber_invoice_status_transitions' is too long; the limit is 64 characters
```

we should suggest shorter name

### DIFF
--- a/lib/generators/active_record_doctor/add_indexes/add_indexes_generator.rb
+++ b/lib/generators/active_record_doctor/add_indexes/add_indexes_generator.rb
@@ -56,7 +56,14 @@ MIGRATION
     end
 
     def add_index(table, column)
-      "    add_index :#{table}, :#{column}"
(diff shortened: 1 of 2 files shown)


gregnavis:
@duleorlovic, thank you for your contribution.

Is it possible to ask Rails for the index name instead of determining it on our own? This would make the code more future-proof as supporting multiple versions of Rails can be quite painful.

duleorlovic:
@gregnavis I found [index_name](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L903) so I updated to use that method.

gregnavis:
Nice work, @duleorlovic! I added a test cases for the generator today and took the liberty of rebasing your branch and adding a test case for your changes.

Thank you for your contribution!

PR #56, opened 2021-04-01
Add a missing migration version to the generated migration file for Rails 6.0

tatsuyafw:
In Rails 6.0, `rails generate active_record_doctor:add_indexes` command generates a migration file without a migration version:

```
$ bin/rails --version
Rails 6.0.3.6
$ bin/rails generate active_record_doctor:add_indexes unindexed_foreign_keys.txt
$ cat db/migrate/20210401171205_index_foreign_keys_in_books.rb
class IndexForeignKeysInBooks < ActiveRecord::Migration
  def change
    add_index :books, :author_id
  end
end
$
```

This patch fixes the issue.

### DIFF
--- a/lib/generators/active_record_doctor/add_indexes/add_indexes_generator.rb
+++ b/lib/generators/active_record_doctor/add_indexes/add_indexes_generator.rb
@@ -54,10 +54,9 @@ EOF
     end
 
     def migration_version
-      major = ActiveRecord::VERSION::MAJOR
-      minor = ActiveRecord::VERSION::MINOR
-      if major >= 5 && minor >= 1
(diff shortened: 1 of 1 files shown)


gregnavis:
@tatsuyafw, thank you for your contribution. Your patch is now on `master` in XXX. I took the liberty of making a slight change there - using floating point isn't advisable due to precision loss. I changed it to a string comparison. Your authorship of the fix was retained, naturally.

I'm closing the PR but if you have any questions or comments then feel free to post them here and/or reopen.

tatsuyafw:
@gregnavis 

Thanks for your comment!

PR #57, opened 2021-05-05
work with ActiveRecord::VERSION::MAJOR >= 5 (not just == 5)

vanharen:
was using Active Record Doctor with a Rails 6 project and ran into a snag -- it was checking for Rails == 5.  simply changed this check to >= 5 in order to get it to work with Rails 6.

### DIFF
--- a/lib/active_record_doctor/tasks/base.rb
+++ b/lib/active_record_doctor/tasks/base.rb
@@ -31,7 +31,7 @@ module ActiveRecordDoctor
 
       def tables
         @tables ||=
-          if ActiveRecord::VERSION::MAJOR == 5
+          if ActiveRecord::VERSION::MAJOR >= 5
             connection.data_sources
           else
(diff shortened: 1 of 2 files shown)


gregnavis:
@vanharen, I'm closing this PR as after the refactoring we no longer use `#data_sources`. However, your PR made me aware of the problem and I'll make sure you're mentioned in the change log.

vanharen:
sorry i didn't get back to you sooner -- i glanced at your change and it
looks good but i have to admit i have not had time to test it out, sorry!

i'll pick up the new gem from rubygems when it gets pushed and play with it
then.

i'm not worried about attribution but if you feel like adding my name,
that's cool.  i didn't really do any work so i don't feel like i
deserve any credit!  :-)

thanks for the email.  i'll definitely keep you posted once i try out the
new version.

-C.

PR #58, opened 2021-05-17
Detect short integer types for primary and foreign keys

### LINKED ISSUE #38: Suggestion: Check for IDs using Integers instead of BigInt
Given no one is immune to it https://m.signalvnoise.com/update-on-basecamp-3-being-stuck-in-read-only-as-of-nov-8-922am-cst/

and there's a lot of people using Rails from prior this https://github.com/rails/rails/pull/26266

So a check for it would be awesome!

fatkodima:
Closes #38 

This checks for both primary keys and foreign keys. Readme wording probably can be improved.

Not related to this PR: we currently can only run one check at a time. Any plans to support running all checks at once by single command? Can work on that.


gregnavis:
Thank you for the contribution, @fatkodima!

I think it'd be superior to split this check into two:

1. Check whether primary keys use short integer types.
2. Check whether primary keys and foreign keys use the same type.

The reason is that primary and foreign keys should use the same type irrespective of what that type is - short int, long int, UUID, string, whatever. If those types are incompatible then I think it's a potential bug.

The primary key check would then be able to identify tables that use the wrong type.

Also, do you think we'd benefit from a migration generator to clean up such issues?

> Not related to this PR: we currently can only run one check at a time. Any plans to support running all checks at once by single command? Can work on that.

It's already there - `active_record_doctor:all`.

fatkodima:
Updated with two separate checks.

Regarding migrations generators: 
1) for mismatched types, I think, it is better to delegate this to the user, because we can only guess which type to use: from the primary key or from the foreign key
2) for primary keys probably a good idea, but also we can only guess or assume, for example, standard type for newest railses - `bigint`

Right now, many more existing checks can be improved with autogenerated migrations and there is a suggestion in https://github.com/gregnavis/active_record_doctor/issues/63 to make it easier. So can add suggested migrator generators as part of that feature.

Something like
```
bundle exec rake active_record_doctor:missing_indexes --migration
``` 
and this will generate the migration, if check supports it, and user only needs to review/edit it. Instead of several manual steps, like here https://github.com/gregnavis/active_record_doctor#indexing-unindexed-foreign-keys

fatkodima:
Updated with suggestions. 
Sure, you can make edits to README. I know, I'm pretty bad at writing documentation 😄

gregnavis:
Thanks for the updates.

I made no comments regarding your documentation skills! My primary intent is to go over README to ensure consistent and smooth style. I'll try to do that today.

PR #60, opened 2021-05-19
Allow usage with non-Rails projects

fatkodima:


gregnavis:
Thank you for the contribution. I'd like to share some thoughts/observations. @fatkodima, please **do not proceed with development** before we decide how to address this problem. I wouldn't like to waste your time on something that won't get merged and I'm not 100% sure what the right approach here is. Let's discuss it first.

### Preloading Models[

There's seems to be a discrepancy between Rails and non-Rails project when it comes to module loading. For Rails, we `eager_load!` in Rake tasks. For non-Rails, we expect the user to load the models. I think it'd be ideal to have a single entry-point for module loading in active_record_doctor. For example:

```ruby
module ActiveRecordDoctor
  def self.load_models
    if defined?(Rails)
      Rails.application.eager_load!
    else
      # We need to come up with some mechanism for loading non-Rails models.
      load "..."
    end
  end
end
```

We could then call `load_models` from `Task` so that project models would always be loaded no matter how `Task` was invoked (via Rake or via CLI).

### Command-Line Interface

Ideally, active_record_doctor should offer a single, consistent user interface for all projects. Currently, there would be a difference between Rails and non-Rails. I think having a standalone binary called `active_record_doctor` could be a good idea. We could also offer `ActiveRecordDoctor::RakeTask` similarity to MiniTest and Spec.

Below is a diagram illustrating the ideas:

test

fatkodima:
Agreed, your suggestions lead to more unified experience working with this feature.

1. I was wondering from the beginning, why it was implemented through separate independent rake tasks and not by one `bin/...` executable. Is that due to the lack of ability to configure this gem yet to skip checks, for example, you do not want?
2. We currently have `:all` rake tasks and kinds of `:one_separate_check` for each check. And we will additionally introduce `bin/...` file and `ActiveRecordDoctor::RakeTask`. 
They are overlapping and can be confusing to decide which to use when. For example, for CI in my project's `Rakefile` should I use `task: [:test, :rubocop, 'active_record_doctor:all']` or  `ActiveRecordDoctor::RakeTask`?
3. > # We need to come up with some mechanism for loading non-Rails models.
How about a config option for `ActiveRecordDoctor::RakeTask` and for `bin/active_record_doctor` to specify the file path to read that will load all models?

gregnavis:
Re 1: The project started as a Rails plugin and I simply used whatever the Rails generator created.

Re 2: I think the way forward should be `bin/active_record_doctor` with an option to declare a Rake task (similarly to how RSpec or Rubocop do it). If that's the goal we should come up with a migration path. I was thinking about:

1. Add `bin/active_record_doctor`.
2. Add `ActiveRecordDoctor::RakeTask`.
3. Keep `active_record_doctor.rake` but make it use the task from step 2. This file is useful for Rails projects because it's auto-included. Non-Rails projects that would like to use a Rake task would need to manually add `ActiveRecordDoctor::RakeTask` to `Rakefile`.

Re 3: I'd like to choose a sensible default so that even if someone runs `bin/active_record_doctor` then it'd do the right thing. Another concern is I'm working on a refactoring that would make detectors configurable (e.g. ignoring errors, skipping tables, etc.). This configuration has to live somewhere too. I'm a bit worried about creating **two** files -- one for loading modules and one for configuration -- as this would pollute the project root directory. I'm currently thinking about adding `.active_record_doctor.rb` that would look like:

```ruby
load_models do
  # ...
end

config do
  # ...
end
```

Basically, it'd be a DSL for configuring Active Record Doctor.

How does that sound?

fatkodima:
@gregnavis 
So, after https://github.com/gregnavis/active_record_doctor#additional-installation-steps-for-non-rails-projects, I think this PR is no longer needed?

gregnavis:
Yes, I think so, @fatkodima. Thank you for your contribution, though. 🙇🏻‍♂️

PR #61, opened 2021-05-20
Enhance IncorrectDependentOption to check belongs_to associations

fatkodima:
`belongs_to` also has `:dependent` option.

And a new case for `belongs_to` is added to this check class, solving the following:
```ruby
class Company < ApplicationRecord
  has_many :users
  has_many :projects
end

class User < ApplicationRecord
  belongs_to :company, dependent: :delete
end

class Project < ApplicationRecord
  belongs_to :company
end
```

If we run `user.destroy`, it will also run `DELETE` for related `company`, and leave related `project`s orphaned. 

The case is also mentioned in documentation - https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-dependent

### DIFF
--- a/lib/active_record_doctor/detectors/incorrect_dependent_option.rb
+++ b/lib/active_record_doctor/detectors/incorrect_dependent_option.rb
@@ -4,7 +4,7 @@ require "active_record_doctor/detectors/base"
 
 module ActiveRecordDoctor
   module Detectors
-    # Find has_many/has_one associations with dependent options not taking the
+    # Find has_many/has_one/belongs_to associations with dependent options not taking the
     # related model's callbacks into account.
     class IncorrectDependentOption < Base
       # rubocop:disable Layout/LineLength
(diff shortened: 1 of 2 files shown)


gregnavis:
I think the check needs to be a bit more complicated than that in order to be reliable. We'd like to inform the user only if there's a real problem. However, with the current implementation, if `users.company_id` **and** `projects.company_id` are defined with `on_delete: :cascade` (at the database level) then it is safe to use `dependent: :delete` on `users`.

It seems the algorithm should be more sophisticated in order to give reliable error messages. I'll do some research and will post notes here.

Also, keep in mind that there's probably a risk of skipping callbacks on `Company` and we should take that into account.

gregnavis:
I think there's a nice recursive algorithm to determine whether it should be `:delete` or `:destroy`.

Let's go through your example: I call `user.destroy`. The callback defined by `dependent: :delete` calls `user.project.delete`. We need to answer the question: when is it safe to delete a model?

The answer in pseudo-code is this:

```ruby
# In this case Model will be Company **NOT** User.
def deletable?(model)
  # We can delete a model without callbacks as they wouldn't be called.
  !defines_destroy_callbacks?(model) &&
    dependent_models(model).all? do |dependent_model|
    end
end
```

gregnavis:
I think there's a nice recursive algorithm to determine whether it should be `:delete` or `:destroy`.

Let's go through your example: I call `user.destroy`. The callback defined by `dependent: :delete` calls `user.project.delete`. We need to answer the question: when is it safe to delete a model?

The answer in pseudo-code is this:

```ruby
# In this case Model will be Company **NOT** User.
def deletable?(model)
  # We can delete a model without callbacks as they wouldn't be called and ...
  !defines_destroy_callbacks?(model) &&
    # for all dependent models (User and Project in the example) ...
    dependent_models(model).all? do |dependent_model|
      # they use NULLIFY or ...
      dependent_model.foreign_key.dependent == :nullify || (
        # or they cascade the deletion AND they are deletable themselves
        dependent_model.foreign_key.dependent == :cascade && deletable?(dependent_model)
      )
      # If dependent_model uses :restrict or nil (NO ACTION) then it's not deletable.
    end
end
```

What do you think?

fatkodima:
Updated with your suggestion. Only CI needs to be fixed for rails 4.2. Can you, please, review?

PR #62, opened 2021-05-20
Rework testing rake tasks

fatkodima:
I think, this is a small, but still valuable improvement. 
I personally forgot several times that we always need to set `DATABASE` variable when testing, and many more times if this should be "postgresql" or just "postgres"? 😄

Maybe even better as:
```ruby
when /postgres/i
  # ...
when /mysql/i
  # ...
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,14 +20,23 @@ Bundler::GemHelper.install_tasks
 
 require "rake/testtask"
 
-Rake::TestTask.new(:test) do |t|
-  t.libs = ["lib", "test"]
-  t.ruby_opts = ["-rsetup"]
-  t.pattern = "test/**/*_test.rb"
-  t.verbose = false
+namespace :test do
+  ["postgresql", "mysql2"].each do |adapter|
+    Rake::TestTask.new(adapter) do |t|
+      t.deps = ["set_#{adapter}_env"]
+      t.libs = ["lib", "test"]
+      t.ruby_opts = ["-rsetup"]
+      t.pattern = "test/**/*_test.rb"
+      t.verbose = false
 
-  # Hide warnings emitted by our dependencies.
-  t.warning = false
+      # Hide warnings emitted by our dependencies.
+      t.warning = false
+    end
+
+    task("set_#{adapter}_env") { ENV["ADAPTER"] = adapter }
+  end
 end
+
+task test: ["test:postgresql", "test:mysql2"]
 
 task default: :test
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ test/dummy/log/*.log
 test/dummy/tmp/
 test/dummy/.sass-cache
 *.gem
+Gemfile.lock
+gemfiles/**.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ os: linux
 dist: xenial
 language: ruby
 
-env:
-  - DATABASE=postgresql
-  - DATABASE=mysql
-
 rvm:
   # Supporting Ruby 2.1+ covers almost the whole market.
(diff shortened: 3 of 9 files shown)


gregnavis:
I'm not sure about picking a default. Specifying the database makes it explicit what you're testing against. I find this helpful given work on this project isn't full-time so it's easy to forget the defaults.

I can see the following alternative:

1. Make database-specific test cases `test:mysql` and `test:postgresql`.
2. Make `test` run them **both**.

Also, we're currently using two variables - `DATABASE` and `DATABASE_URL`. Theoretically it's possible to set one to `postgres` and the other one to `mysql`. Maybe we should rely on `DATABASE` only?

fatkodima:
Yes, makes sense. 
Suggested alternative looks great to me and this is the way how [`active_record`s own tests](https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#testing-active-record) against multiple adapters are implemented.

Will update this PR with this suggestion.

fatkodima:
Updated.

Rakefile
```diff
@@ -20,14 +20,23 @@ Bundler::GemHelper.install_tasks
...
-  t.libs = ["lib", "test"]
-  t.ruby_opts = ["-rsetup"]
-  t.pattern = "test/**/*_test.rb"
-  t.verbose = false
-
-  # Hide warnings emitted by our dependencies.
-  t.warning = false
+namespace :test do
+  ["postgresql", "mysql2"].each do |adapter|
+    adapter_short = adapter[/^[a-z]+/]
```
gregnavis, line 25: Given it's not simple to run the test suite against multiple database, I suggest we remove the distinction between `mysql` and `mysql2` by using the adapter name (`mysql2`) - this way we'll remove code mapping from one to the other thus reducing the overall cognitive overhead.

Rakefile
```diff
@@ -20,14 +20,23 @@ Bundler::GemHelper.install_tasks
...
+    Rake::TestTask.new(adapter_short => "set_#{adapter_short}_env") do |t|
+      t.libs = ["lib", "test"]
+      t.ruby_opts = ["-rsetup"]
+      t.pattern = "test/**/*_test.rb"
+      t.verbose = true
```
gregnavis, line 30: Seems the value's flipped from `false` in the original code to `true` here. Is that intentional?

Rakefile
```diff
@@ -20,14 +20,23 @@ Bundler::GemHelper.install_tasks
...
-  t.pattern = "test/**/*_test.rb"
-  t.verbose = false
-
-  # Hide warnings emitted by our dependencies.
-  t.warning = false
+namespace :test do
+  ["postgresql", "mysql2"].each do |adapter|
+    adapter_short = adapter[/^[a-z]+/]
+    Rake::TestTask.new(adapter_short => "set_#{adapter_short}_env") do |t|
+      t.libs = ["lib", "test"]
```
gregnavis, line 27: I'd stick to the public API of `Rake::TestTask` which indicates the constructor expects `name` as the only parameter. We can pass deps via `t.deps = ...`.

gregnavis (changes requested):
Looks great! I left some comments about mostly cleaning things up a bit.

Thanks for the contribution!

Rakefile
fatkodima: Than seems like a public API, I think - https://github.com/ruby/rake/blob/fc688bd8cd6b4178178a0d7bb31068122ba25943/lib/rake/testtask.rb#L98-L101

But nevertheless, changed to `t.deps = ...`

fatkodima:
Thank you for comments! Updated with suggestions.

gregnavis:
@fatkodima, merged! 🚀

PR #68, opened 2021-08-31
Ignore automatically generated HABTM models

### LINKED ISSUE #67: missing_presence_validation: unexpected error on HABTM join tables
Hi!

when running `rake active_record_doctor:missing_presence_validation`, the gem reports a missing validation error on the FK columns on join tables (used in HABTM relations).

Is this really the intended behavior of this gem? Since there is no model for the join table, it shouldn't report an error IMHO.

Example with schema:
```rb
  create_table "pets", force: :cascade do |t|
    t.string "name", null: false
    t.date "birth", null: false
  end

  create_table "pets_users", id: false, force: :cascade do |t|
    t.integer "users_id", null: false
    t.integer "pets_id", null: false
    t.index ["pets_id"], name: "index_pets_users_on_pets_id"
    t.index ["users_id"], name: "index_pets_users_on_users_id"
  end

  create_table "users", force: :cascade do |t|
    t.string "email"
  end
```

Pet:
```rb
  class Pet < ApplicationRecord
    has_and_belongs_to_many :users
    validates :name, presence: true
    validates :birth, presence: true
  end
```

User model
```rb
class User < ApplicationRecord
  has_and_belongs_to_many :pets
end
```

Output:
```
The following models and columns should have presence validations:
  HABTM_Users: users_id, pets_id
  HABTM_Pets: users_id, pets_id
```

Expected output: no error

* gem version: ` active_record_doctor (1.8.0)`

fatkodima:
Rails' `has_and_belongs_to_many` automatically [creates](https://github.com/rails/rails/blob/ce546c7d3a867dce11e7b99624160a7388bee1f5/activerecord/lib/active_record/associations/builder/has_and_belongs_to_many.rb#L14) intermediate models, which should really not be checked.

Closes #67

### DIFF
--- a/lib/active_record_doctor/detectors/base.rb
+++ b/lib/active_record_doctor/detectors/base.rb
@@ -87,7 +87,8 @@ module ActiveRecordDoctor
       end
 
       def models
-        ActiveRecord::Base.descendants
+        # Ignore auto-generated models by Active Record has_and_belongs_to_many macro
+        ActiveRecord::Base.descendants.reject { |model| model.name.start_with?("HABTM_") }
       end
     end
   end
--- a/test/active_record_doctor/detectors/missing_presence_validation_test.rb
+++ b/test/active_record_doctor/detectors/missing_presence_validation_test.rb
@@ -42,6 +42,31 @@ OUTPUT
     refute_problems
   end
 
+  def test_not_null_column_is_not_reported_if_habtm_association
+    create_table(:users).create_model do
+      def self.name
+        "ModelFactory::Models::User"
+      end
+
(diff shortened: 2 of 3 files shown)


gregnavis:
Thank you for your contribution, @fatkodima. I merged the change in d24c96cc371f2a546fd997b86f69981611ebd168 but made some changes. You can have a look here: https://github.com/gregnavis/active_record_doctor/commit/d24c96cc371f2a546fd997b86f69981611ebd168#diff-0752ba99cd7a3f6acddfe0dd59381754dc45962f414e2445f6e2fa0f5bbc3f43R91-R110

PR #70, opened 2021-11-11
Fix ExtraneousIndexes to ignore incompatible indexes

### LINKED ISSUE #69: False positive for extraneous Postgres trigram index
Here is a fragment from my `schema.rb` file that is flagged by `rails active_record_doctor:extraneous_indexes` as having an extraneous index that I believe is a false positive.

```ruby
t.index ["code", "account_id"], name: "index_skus_on_code_and_account_id", unique: true
t.index ["code"], name: "index_skus_on_code_trgm", opclass: :gin_trgm_ops, using: :gin
```
The output from running `rails active_record_doctor:extraneous_indexes` is:

```text
The following indexes are extraneous and can be removed:
  index_skus_on_code_trgm (can be handled by index_skus_on_code_and_account_id)
```

Thanks for making such a useful tool.

fatkodima:
Closes #69

### DIFF
--- a/lib/active_record_doctor/detectors/extraneous_indexes.rb
+++ b/lib/active_record_doctor/detectors/extraneous_indexes.rb
@@ -70,6 +70,8 @@ module ActiveRecordDoctor
 
       # Does lhs cover rhs?
       def cover?(lhs, rhs)
+        return false unless compatible_options?(lhs, rhs)
+
         case [lhs.unique, rhs.unique]
         when [true, true]
           lhs.columns == rhs.columns
@@ -87,6 +89,25 @@ module ActiveRecordDoctor
 
       def indexes(table_name)
         super.select { |index| index.columns.is_a?(Array) }
+      end
+
+      def compatible_options?(lhs, rhs)
+        lhs.type == rhs.type &&
+          lhs.using == rhs.using &&
+          lhs.where == rhs.where &&
+          same_opclasses?(lhs, rhs)
+      end
+
+      def same_opclasses?(lhs, rhs)
+        if ActiveRecord::VERSION::STRING >= "5.2"
+          rhs.columns.all? do |column|
+            lhs_opclass = lhs.opclasses.is_a?(Hash) ? lhs.opclasses[column] : lhs.opclasses
+            rhs_opclass = rhs.opclasses.is_a?(Hash) ? rhs.opclasses[column] : rhs.opclasses
+            lhs_opclass == rhs_opclass
+          end
+        else
+          true
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


test/active_record_doctor/detectors/extraneous_indexes_test.rb
```diff
@@ -61,6 +61,46 @@ def test_multi_column_covered_by_unique_and_non_unique_multi_column_is_duplicate
 OUTPUT
   end
 
+  def test_not_covered_by_different_index_type
+    create_table(:users) do |t|
+      t.string :first_name
+      t.string :last_name
+      t.index [:last_name, :first_name], using: :btree
+      t.index :last_name, using: :hash
```
fatkodima, line 69: While technically the second index is covered by the first, I think, originally `ExtraneousIndexes` check was meant to work with B-trees only, as this could cover 99% of the cases and makes its implementation simpler.

So I limited it's checks only for B-tree indexes and assume differing index types as incompatible.

lib/active_record_doctor/detectors/extraneous_indexes.rb
```diff
@@ -41,9 +41,10 @@ def subindexes_of_multi_column_indexes
             next if maximal_indexes.include?(index)
 
             replacement_indexes = maximal_indexes.select do |maximum_index|
-              cover?(maximum_index, index)
+              compatible_options?(maximum_index, index) && cover?(maximum_index, index)
             end.map(&:name).sort
 
+            next if replacement_indexes.empty?
```
gregnavis, line 47: These changes must be reverted and the `#cover?` method must be patched instead. Let me explain the thinking behind this algorithm.

Indexes form a partially ordered set - you can visualize them using directed graphs or Hasse diagrams. Indexes are vertices in a directed graph and there's an edge from index `a` to index `b` if `a` can replace `b`. In the code we call it `#cover?` so if `cover?(a, b)` returns `true` then `a` covers (can replace) `b`.

The problem is `#cover?` doesn't check for operator classes. Currently, it takes uniqueness into account. We must modify `#cover?` to take operator classes into account so that when it receives two indexes with different types or different operator classes then it always returns `false`. After doing that, the algorithm should do just fine without any changes to `subindexes_of_multi_column_indexes`.

lib/active_record_doctor/detectors/extraneous_indexes.rb
```diff
@@ -86,7 +87,25 @@ def prefix?(lhs, rhs)
       end
 
       def indexes(table_name)
-        super.select { |index| index.columns.is_a?(Array) }
+        super.select { |index| index.columns.is_a?(Array) && index.using == :btree }
```
gregnavis, line 90: We should be able to revert this after we fix `#cover?` as explained above.

gregnavis (changes requested):
Thanks for the PR, @fatkodima. I left some comments explaining how to fix that bug within the theoretical framework behind the algorithm. Let me know if you have any questions.

fatkodima:
Thank you for the review! Updated. 
Agreed, this is the correct way.

PR #71, opened 2021-11-11
Reduce gemfiles to depend on activerecord

fatkodima:
No need to install a lot of extra dependencies.


gregnavis:
Merged! Thank you, @fatkodima!

Issue #65, opened 2021-06-10
Check for has_one backed by a unique index


fatkodima:
```ruby
class User
  has_one :account
end

class Account
  belongs_to :user
end
```

`has_one` should be backed by a unique index (on `accounts.user_id`), because otherwise user can have multiple accounts in the database.

There is already a check `MissingUniqueIndexes`, checking unique validations without indexes. It is pretty generically named and seems like this new check would be added into there? Or a new check created?

@gregnavis, wdyt?

fatkodima:
@gregnavis Is this a good idea?

gregnavis:
@fatkodima, I think it's a great idea! `MissingUniqueIndexes` seems to be the right place for that check.

PR #81, opened 2022-01-11
Check for has_ones without unique indexes

fatkodima:
Closes https://github.com/gregnavis/active_record_doctor/issues/65

### DIFF
--- a/lib/active_record_doctor/detectors/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/detectors/missing_unique_indexes.rb
@@ -18,17 +18,23 @@ module ActiveRecordDoctor
 
       private
 
-      def message(table:, columns:)
-        # rubocop:disable Layout/LineLength
-        "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
-        # rubocop:enable Layout/LineLength
+      # rubocop:disable Layout/LineLength
+      def message(model:, table:, columns:, problem:)
+        case problem
+        when :validations
+          "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
+        when :has_ones
+          "add a unique index on #{table}(#{columns.first}) - using `has_one` in the #{model.name} model without an index can lead to duplicates"
+        end
       end
+      # rubocop:enable Layout/LineLength
 
       def detect
-        ignore_columns = config(:ignore_columns).map do |column|
-          column.gsub(" ", "")
-        end
+        validations_without_indexes
+        has_ones_without_indexes
+      end
 
+      def validations_without_indexes
         models(except: config(:ignore_models)).each do |model|
           next unless model.table_exists?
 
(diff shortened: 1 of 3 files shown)


lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -18,17 +18,23 @@ class MissingUniqueIndexes < Base # :nodoc:
...
-        # rubocop:disable Layout/LineLength
-        "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
-        # rubocop:enable Layout/LineLength
+      # rubocop:disable Layout/LineLength
+      def message(model:, table:, columns:, problem:)
+        case problem
+        when :validations
+          "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
+        when :has_ones
+          "add a unique index on #{table}(#{columns.first}) - using `has_one` in the #{model.name} model without an index can lead to duplicates"
```
gregnavis, line 27: I suggest we use `columns.join(', ')` - the result will be the same for single-entry arrays but we'll immediately see something is wrong in case we pass a larger array in a future version.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -42,7 +48,27 @@ def detect
...
+      end
+
+      def has_ones_without_indexes # rubocop:disable Naming/PredicateName
+        models.each do |model|
+          has_ones = model.reflect_on_all_associations(:has_one)
+          has_ones.each do |has_one|
+            next if has_one.is_a?(ActiveRecord::Reflection::ThroughReflection) || has_one.scope
+
+            association_model = has_one.klass
+            next if config(:ignore_models).include?(association_model.name)
```
gregnavis, line 63: I think I understand your intent here. However, using `ignore_models` like that deviates from how it's used in all other detectors and might be confusing to users: adding a model to the ignored list would ignore its validations but **not** it's `has_one` associations.

I suggest we use `ignore_models` like in all other detectors (i.e. `models(except: config(:ignore_models)).each ...`). If the user wants to silence a warning he should use `ignore_columns` - that would be consistent with how it works in the case of validations. When the users sees:

```
add a unique index on users(organization_id) - ...
```

then he knows this can be ignored by adding `"users.organization_id"` to `ignored_columns`.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -59,13 +85,19 @@ def supported_validator?(validator)
...
         indexes(table_name).any? do |index|
           Array(index.columns).to_set == columns.to_set && index.unique
         end
       end
+
+      def ignore_columns
+        @ignore_columns ||= config(:ignore_columns).map do |column|
+          column.gsub(" ", "")
+        end
+      end
```
gregnavis, line 100: Why do we need this?

test/.rubocop.yml
```diff
@@ -11,3 +11,6 @@ Style/ClassAndModuleChildren:
 # Some test models need methods defined on them in a #create_model block.
 Lint/NestedMethodDefinition:
   Enabled: false
+
+Layout/LineLength:
+  Enabled: false
```
gregnavis, line 16: I'd prefer to keep this rule enabled and disable it only for user-facing messages with as few other exception as possible. Longer lines are harder to read.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -109,6 +109,57 @@ def test_skips_validator_without_attributes
     refute_problems
   end
 
+  def test_has_one_without_index
```
gregnavis, line 112: We also need three other test cases:

- `ignore_models` defined globally
- `ignore_models` defined on the detector only
- `ignore_columns` defined on the detector (there's no global setting)

You can have a look at other test cases for reference.

gregnavis (changes requested):
@fatkodima, thank you for your contribution!

I rebased the PR on top of the most recent `master` and left some comments. You can address my comments, if you're so inclined, or let me know if you'd like me to do the changes.

gregnavis:
I'm fixing the build failure caused by non-deterministic output line ordering.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
fatkodima: That was moved from `detect` method. I think, it is needed to be able to work with `User(organization_id, email, role)` and `User(organization_id,email,role)` (no spaces) the same way.

fatkodima:
Updated. Let me know what you think or there is a need to squash/rebase.

stevenhan2:
Looks useful

gregnavis (approved):

PR #84, opened 2022-02-21
Better support for PostGIS

fatkodima:
https://github.com/rgeo/activerecord-postgis-adapter/blob/4f4e28fa3c21f4b31b6bdaea484a3872fe3af252/lib/active_record/connection_adapters/postgis_adapter.rb#L27

### DIFF
--- a/lib/active_record_doctor/detectors/base.rb
+++ b/lib/active_record_doctor/detectors/base.rb
@@ -131,7 +131,7 @@ module ActiveRecordDoctor
         @views ||=
           if connection.respond_to?(:views)
             connection.views
-          elsif connection.adapter_name == "PostgreSQL"
+          elsif postgresql?
             ActiveRecord::Base.connection.execute(<<-SQL).map { |tuple| tuple.fetch("relname") }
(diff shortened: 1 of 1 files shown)


lib/active_record_doctor/detectors/base.rb
```diff
@@ -150,6 +150,10 @@ def models(except: [])
       def underscored_name
         self.class.underscored_name
       end
+
+      def postgresql?
+        connection.adapter_name =~ /postg/i
```
gregnavis, line 155: Let's use an array of adapter names that correspond to PostgreSQL. Less matching adapters, higher guarantee those that do match actually work.

gregnavis (changes requested):
Thank you for the contribution. I left one comment.

lib/active_record_doctor/detectors/base.rb
fatkodima: Changed.

gregnavis: Thank you. I'll merge tomorrow.

gregnavis:
Merged in 92b9479ae00125bf48f8027bb39ffe78e1134551. Thank you, @fatkodima!

PR #86, opened 2022-03-03
Fix typo in Missing Presence Validations config options

pingortle:
It looks like the correct configuration option according to the tests is `ignore_attributes`.

Source:

https://github.com/gregnavis/active_record_doctor/blob/32906dc95f4343290c237026596a535c7348351f/test/active_record_doctor/detectors/missing_presence_validation_test.rb#L153-L168


gregnavis:
Thank yoiu, @pingortle.

PR #89, opened 2022-03-05
Fix `MissingUniqueIndexes` detector for functional indexes

### LINKED ISSUE #82: undefined method `to_set' for String, when using a lowered index
My migration is:

```ruby
add_index :users, 'lower(email)', unique: true
```

It translates in `schema.rb` as:

```ruby
t.index "lower((email)::text)", name: "index_users_on_lower_email", unique: true
```

Note there are no surrounding brackets like we have for other fields:

```ruby
add_index :users,  :auth_token, unique: true
```

```ruby
t.index ["auth_token"], name: "index_users_on_auth_token", unique: true
```

The error is:

```
NoMethodError: undefined method `to_set' for "lower((email)::text)":String

          index.columns.to_set == columns.to_set && index.unique
```

in `active_record_doctor-1.9.0/lib/active_record_doctor/detectors/missing_unique_indexes.rb:66:in block in unique_index?`

This can be fixed by making sure `index.columns` is wrapped into an Array:

```ruby
  def unique_index?(table_name, columns, scope)
    columns = (Array(scope) + columns).map(&:to_s)

    indexes(table_name).any? do |index|
      Array(index.columns).to_set == columns.to_set && index.unique
    end
  end
```

fatkodima:
Closes #82

### DIFF
--- a/lib/active_record_doctor/detectors/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/detectors/missing_unique_indexes.rb
@@ -63,7 +63,7 @@ module ActiveRecordDoctor
         columns = (Array(scope) + columns).map(&:to_s)
 
         indexes(table_name).any? do |index|
-          index.columns.to_set == columns.to_set && index.unique
+          Array(index.columns).to_set == columns.to_set && index.unique
         end
       end
     end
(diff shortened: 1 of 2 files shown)


Sokre95:
@fatkodima Thanks for solving this issue. Is there any info that this may be merged soon ?

fatkodima:
@gregnavis is an active maintainer, so I'm sure this will be done very soon.

PR #91, opened 2022-03-12
Ignore abstract model classes

fatkodima:
```bash
$ rake active_record_doctor
```

Got this warning, among others:
```
ApplicationRecord references a non-existent table or view named
```


gregnavis:
I think those errors were introduced after we removed a check for `model.table_name.nil?`. We probably should reintroduce that check.

fatkodima:
> I think those errors were introduced after we removed a check

Double checked - yes. But I think, this PR should cover that case. Any reason to have a model without `table_name`, except abstract classes?

PR #97, opened 2022-05-23
Enhance `MissingUniqueIndexes` to accept unique indexes with sublist of columns

fatkodima:
If we have `validates_uniqueness_of :a, scope: [:b, :c]` it does not require a unique index consisting of strictly `(:a, :b, :c)` columns, but any smaller combination of these (like `(:a)` or `(:a, :c)` etc) works too.

Also this PR adds a check for partial indexes, since they are not guaranteed to always provide uniqueness to a list of columns.


gregnavis:
This is a *great* observation. Thank you for the contribution. Merging!

PR #98, opened 2022-06-05
Enhance `MissingUniqueIndexes` to work correctly with associations and multi-attributes validations

fatkodima:
This PR improves `MissingUniqueIndexes` in a several ways:
1. correctly handles associations in `validates`. Before, it was assumed, that only  column names are specified.
2. ignores partial indexes, since they do not cover uniqueness validation
3. correctly works with multi-attribute validations (`validates :email, :ref_token, uniqueness: true`). Before, it was assumed that all attributes are validated simultaneously and  should be present in a single index, but actually each attribute validates individually.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/detectors/missing_unique_indexes.rb
@@ -38,12 +38,14 @@ module ActiveRecordDoctor
             next unless validator.is_a?(ActiveRecord::Validations::UniquenessValidator)
             next unless supported_validator?(validator)
 
-            columns = (scope + validator.attributes).map(&:to_s)
-            next if unique_index?(model.table_name, columns)
+            validator.attributes.each do |attribute|
+              columns = resolve_attributes(model, scope + [attribute])
 
-            next if ignore_columns.include?("#{model.name}(#{columns.join(',')})")
+              next if unique_index?(model.table_name, columns)
+              next if ignore_columns.include?("#{model.name}(#{columns.join(',')})")
 
-            problem!(table: model.table_name, columns: columns)
+              problem!(table: model.table_name, columns: columns)
+            end
           end
         end
       end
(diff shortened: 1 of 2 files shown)


gregnavis:
@fatkodima, would you be able to update this PR on top of the current `master`?

fatkodima:
@gregnavis Done.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -60,6 +62,20 @@ def supported_validator?(validator)
...
+          reflection = model.reflect_on_association(attribute)
+
+          if reflection.nil?
+            attribute
+          elsif reflection.polymorphic?
+            [reflection.foreign_key, reflection.foreign_type]
+          else
+            reflection.foreign_key
+          end
+        end.flatten.map(&:to_s)
```
gregnavis, line 76: Are we able to use `flat_map` instead of `map.flatten`?

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -30,6 +30,20 @@ def test_missing_unique_index_on_functional_index
...
+    create_table(:users) do |t|
+      t.string :email
+      t.string :ref_token
+      t.index :email, unique: true
+    end.create_model do
+      validates :email, :ref_token, uniqueness: true
+    end
+
+    assert_problems(<<~OUTPUT)
+      add a unique index on users(ref_token) - validating uniqueness in the model without an index can lead to duplicates
```
gregnavis, line 43: I think the test case should remove the unique index on `email` and test whether both `email` and `ref_token` are reported -- otherwise, we'd let an implementation testing only the first attribute pass through the tests.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -60,6 +62,20 @@ def supported_validator?(validator)
...
       end
 
+      def resolve_attributes(model, attributes)
+        attributes.map do |attribute|
+          reflection = model.reflect_on_association(attribute)
+
+          if reflection.nil?
+            attribute
+          elsif reflection.polymorphic?
+            [reflection.foreign_key, reflection.foreign_type]
```
gregnavis, line 72: I think it'd make more sense to have the type appear first, followed by ID, as that index could also participate in fetching related models (e.g. "fetch all comments left on articles", which could use the first column in an index on `(type, id)`).

gregnavis (changes requested):
Awesome work! Thank you for the contribution, @fatkodima.

I requested a few changes but they're relatively minor.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
gregnavis: Yes, the order doesn't matter for uniqueness but it matters for **other** queries. If we're recommending creating an index then let's recommend something the option with broader practical applications so that the user won't need to create another index on the same columns but in reverse order.

fatkodima: @gregnavis Ahh, sure 🤦 Updated.

gregnavis (approved):

PR #101, opened 2022-06-16
Reduce the number of queries by caching schema info

### LINKED ISSUE #99: Use `SchemaCache` internally
We can use [`SchemaCache`](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaCache.html) internally to not request the same information from the db over and over. I am sure, this will improve runtime on decent sizeable projects with many models/tables.

Unfortunately, I do not have a decent project under my hands right now to test the actual improvement.

fatkodima:
Fixes #99.

I was able to run `active_record_doctor` on 2 OSS projects: `discourse` (medium sized) and `gitlab` (large sized).

### Discourse
**Before**
Num of queries: `6771`
Runtime of `active_record_doctor` (checks only, not including app eager loading): `8.53s`
Total runtime (of the `rake active_record_doctor`): `15.74s` 

**After**
Num of queries: `1739`
Runtime of `active_record_doctor`: `2.55s` 🔥 
Total runtime : `9.3s`

### Gitlab
**Before**
Num of queries: `31499 `
Runtime of `active_record_doctor`: `125.9s`
Total runtime: `2:36.17m` 

**After**
Num of queries: `10697 `
Runtime of `active_record_doctor`: `38.53s` 🔥 🔥 🔥 
Total runtime: `1:13.80m`


fatkodima:
With the new PR from rails - https://github.com/rails/rails/pull/45381, this will be even faster for gitlab: `31s`.

lib/active_record_doctor/schema_cache.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module ActiveRecordDoctor
+  class SchemaCache # :nodoc:
+    attr_reader :connection
```
gregnavis, line 5: It seems this attribute can be made private.

lib/active_record_doctor/schema_cache.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module ActiveRecordDoctor
+  class SchemaCache # :nodoc:
```
gregnavis, line 4: I suggest using a different name here. The reasoning is that the callers of that class shouldn't care whether it's caching or not. Its primary role is schema reflection/introspection with caching as a side benefit.

My suggestion would be to:

- Rename this to `CachingSchemaInspector` (to make the class name indicate what this class is about).
- Rename all occurrences of `schema_cache` to `schema_inspector` (the callers don't need to know about caching).

lib/active_record_doctor/schema_cache.rb
```diff
@@ -0,0 +1,88 @@
...
+    attr_reader :connection
+
+    def initialize(connection)
+      @connection = connection
+      @indexes = {}
+      @foreign_keys = {}
+      @check_constraints = {}
+    end
+
+    def tables
```
gregnavis, line 14: Seems these methods were simply moved without changes (except for caching), but @fatkodima please confirm.

gregnavis (changes requested):

lib/active_record_doctor/schema_cache.rb
fatkodima: Yes.

gregnavis:
@fatkodima, please rebase and it's good to merge!

fatkodima:
@gregnavis Rebased.

fatkodima:
@gregnavis Would you like to also add this to 2.0? I can update it then.

fatkodima:
@gregnavis Merge? Just tested on my app - 2x speedup.

PR #106, opened 2022-06-19
Ignore non-integer and non-uuid primary keys for short_primary_key_type

fatkodima:
It is possible the user uses other types for primary keys, for example string `username` for `users` table, or client side generated uuid-like strings for some other tables, or external ids (like `stripe_id` string) for `stripe_users` table, etc.

### DIFF
--- a/lib/active_record_doctor/detectors/short_primary_key_type.rb
+++ b/lib/active_record_doctor/detectors/short_primary_key_type.rb
@@ -23,7 +23,7 @@ module ActiveRecordDoctor
         each_table(except: config(:ignore_tables)) do |table|
           column = primary_key(table)
           next if column.nil?
-          next if bigint?(column) || uuid?(column)
+          next if !integer?(column) || bigint?(column)
 
           problem!(table: table, column: column.name)
         end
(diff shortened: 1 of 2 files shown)


lib/active_record_doctor/detectors/short_primary_key_type.rb
```diff
@@ -23,7 +23,7 @@ def detect
         tables(except: config(:ignore_tables)).each do |table|
           column = primary_key(table)
           next if column.nil?
-          next if bigint?(column) || uuid?(column)
+          next if bigint?(column) || uuid?(column) || !integer?(column)
```
gregnavis, line 26: I think the problem is with auto-incremented integers, so maybe the check can actually be reduced to:

```ruby
next if !integer?(column) || bigint?(column)
```

What do you think? I think UUIDs don't have to be tackled explicitly.

gregnavis (commented):
@fatkodima, thank you for your contribution. I left one question.

lib/active_record_doctor/detectors/short_primary_key_type.rb
gregnavis: I think the `uuid?` method can now be removed, too.

fatkodima: Ah, oops 💩 Removed.

gregnavis (approved):

PR #107, opened 2022-07-09
Detect missing unique lowercase indexes for case insensitive unique validations

fatkodima:
Case insensitivity in uniqueness validations was skipped in this gem, but this is a very popular problem. 
I remember I was trapped by this at least once. And there is even [a related article from Richard Scheeman](https://schneems.com/2017/07/18/how-i-reduced-my-db-server-load-by-80/) with his investigation of the performance problems on production with his findings.   

It adds some ugly columns parsing, but I think we can deal with it.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/detectors/missing_unique_indexes.rb
@@ -23,6 +23,9 @@ module ActiveRecordDoctor
         case problem
         when :validations
           "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
+        when :case_insensitive_validations
+          "add a unique expression index on #{table}(#{columns.join(', ')}) - validating case-insensitive uniqueness in the model "\
+            "without an expression index can lead to duplicates (a regular unique index is not enough)"
         when :has_ones
           "add a unique index on #{table}(#{columns.first}) - using `has_one` in the #{model.name} model without an index can lead to duplicates"
         end
@@ -40,15 +43,33 @@ module ActiveRecordDoctor
             scope = Array(validator.options.fetch(:scope, []))
 
             next unless validator.is_a?(ActiveRecord::Validations::UniquenessValidator)
-            next unless supported_validator?(validator)
+            next if conditional_validator?(validator)
+
+            # In Rails 6, default option values are no longer explicitly set on
+            # options so if the key is absent we must fetch the default value
+            # ourselves. case_sensitive is the default in 4.2+ so it's safe to
+            # put true literally.
+            case_sensitive = validator.options.fetch(:case_sensitive, true)
+
+            # ActiveRecord < 5.0 does not support expression indexes,
+            # so this will always be a false positive.
+            next if !case_sensitive && Utils.expression_indexes_unsupported?
 
             validator.attributes.each do |attribute|
               columns = resolve_attributes(model, scope + [attribute])
 
-              next if unique_index?(model.table_name, columns)
               next if ignore_columns.include?("#{model.name}(#{columns.join(',')})")
 
-              problem!(model: model, table: model.table_name, columns: columns, problem: :validations)
+              columns[-1] = "lower(#{columns[-1]})" unless case_sensitive
+
+              next if unique_index?(model.table_name, columns)
+
+              if case_sensitive
+                problem!(model: model, table: model.table_name, columns: columns, problem: :validations)
+              else
+                problem!(model: model, table: model.table_name, columns: columns,
+                         problem: :case_insensitive_validations)
+              end
             end
           end
         end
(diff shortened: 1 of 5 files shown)


lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -76,16 +89,8 @@ def has_ones_without_indexes # rubocop:disable Naming/PredicateName
...
-          validator.options[:unless].nil? &&
-          validator.options[:conditions].nil? &&
-
-          # In Rails 6, default option values are no longer explicitly set on
-          # options so if the key is absent we must fetch the default value
-          # ourselves. case_sensitive is the default in 4.2+ so it's safe to
-          # put true literally.
-          validator.options.fetch(:case_sensitive, true)
+      def conditional_validator?(validator)
+        (validator.options.keys & [:if, :unless, :conditions]).any?
```
gregnavis, line 93: I think `present?` would be slightly more appropriate here, as our intent is to test whether the intersection is non-empty, not whether it contains a truthy value.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -116,6 +123,26 @@ def ignore_columns
...
       end
+
+      def expression_indexes_unsupported?
+        (ActiveRecord::VERSION::STRING < "5.0") ||
+          # Active Record < 6 is unable to correctly parse expression indexes for MySQL.
+          (mysql? && ActiveRecord::VERSION::STRING < "6.0")
+      end
+
+      def index_columns(columns)
+        return columns unless columns.is_a?(String)
```
gregnavis, line 134: I'd like to make two points here:

1. Let's add a comment explaining why this is needed. What are types other than `String` that can be passed to the function.
2. Are there cases where this is an array and we'd need to perform the transformations below (those `gsub`s) on each of its elements?

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -116,6 +123,26 @@ def ignore_columns
           column.gsub(" ", "")
         end
       end
+
+      def expression_indexes_unsupported?
+        (ActiveRecord::VERSION::STRING < "5.0") ||
+          # Active Record < 6 is unable to correctly parse expression indexes for MySQL.
+          (mysql? && ActiveRecord::VERSION::STRING < "6.0")
+      end
```
gregnavis, line 131: I suggest extracting this to `ActiveRecordDoctor.expression_indexes_unsupported?` and using it in this detector and the test suite.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -179,12 +179,58 @@ def test_column_order_is_ignored
...
+    create_table(:users) do |t|
+      t.string :email
+
+      t.index :email, unique: true
+    end.create_model do
+      validates :email, uniqueness: { case_sensitive: false }
+    end
+
+    assert_problems(<<~OUTPUT)
+      add a unique index on users(lower(email)) - validating uniqueness in the model without an index can lead to duplicates
```
gregnavis, line 194: I suggest a different error message when the complaint is about a missing **expression** index:

```
add a unique expression index on users(lower(email)) - validating case-insensitive uniqueness in the model without an expression index can lead to duplicates (a regular unique index is not enough)
```

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -179,12 +179,58 @@ def test_column_order_is_ignored
...
+
+    refute_problems
+  end
+
+  def test_case_insensitive_compound_lower_index_exists
+    skip("Expression indexes are not supported") if expression_indexes_unsupported?
+
+    create_table(:users) do |t|
+      t.string :email
+      t.integer :organization_id
```
gregnavis, line 220: I think in this case we also should a unique index on `(organization_id, email)` to ensure only an **expression** index will satisfy the condition. Similarily, we need another test case with a non-unique expression index.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
gregnavis: Also, we need another test case that creates a **non-unique expression index** on the column.

The index must satisfy two conditions to satisfy the detector:

1. Unique
2. Expression

This gives four combinations in total:

1. No index at all (already covered by a previous test case).
2. Unique, no expression (the test case you added).
3. Non-unique, expression (the test case that's missing).
4. Unique, expression (added below).

gregnavis (changes requested):

lib/active_record_doctor/detectors/missing_unique_indexes.rb
fatkodima: 1. Added comment to the callsite
2. Seems like no. Rails returns columns as arrays only when they are "bare" columns. When one of them is an expression, the string is returned

lib/active_record_doctor/detectors/missing_unique_indexes.rb
fatkodima: I extracted a separate module for this kind of utils. It feels not correct to me to pollute public `ActiveRecordDoctor` namespace with such helper functions. Let me know what you think.

fatkodima:
Made suggested changes.

gregnavis (approved):

gregnavis:
Merged! Thank you, @fatkodima.

PR #108, opened 2022-07-15
Check for `dependent: :destroy_async` and a foreign key

fatkodima:
Rails 6.1 [introduced](https://github.com/rails/rails/pull/40157) `dependent: :destroy_async` which allows to destroy associated records in the background. This is useful when the parent record has lots of associated records and destroying all of them in the same transaction can potentially take a lot of time and/or lead to long database locks.

From the [documentation](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many):

> **WARNING**: Do not use this option if the association is backed by foreign key constraints in your database. The foreign key constraint actions will occur inside the same transaction that deletes its owner.  

So one can create a foreign key with `on_delete: :cascade` and an association with `dependent: :destroy_async` and when destroying the parent record, all the child records will be deleted/nullified immediately in the same transaction. This can run unnoticed for some time and blow up production later.

### DIFF
--- a/lib/active_record_doctor/detectors/incorrect_dependent_option.rb
+++ b/lib/active_record_doctor/detectors/incorrect_dependent_option.rb
@@ -18,7 +18,8 @@ module ActiveRecordDoctor
 
       private
 
-      def message(model:, association:, problem:, associated_models:, associated_models_type:)
+      def message(model:, association:, problem:, associated_models_type: nil,
+                  table_name: nil, column_name: nil, associated_models: [])
         associated_models.sort!
 
         models_part =
@@ -36,6 +37,9 @@ module ActiveRecordDoctor
         case problem
         when :invalid_through
           "ensure #{model}.#{association} is configured correctly - #{associated_models[0]}.#{association} may be undefined"
+        when :destroy_async
+          "don't use `dependent: :destroy_async` on #{model}.#{association} or remove the foreign key from #{table_name}.#{column_name} - "\
+            "associated models will be deleted in the same transaction along with #{model}"
         when :suggest_destroy
           "use `dependent: :destroy` or similar on #{model}.#{association} - associated #{models_part} callbacks that are currently skipped"
         when :suggest_delete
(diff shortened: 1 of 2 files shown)


lib/active_record_doctor/detectors/incorrect_dependent_option.rb
```diff
@@ -60,19 +63,29 @@ def detect
...
-
-              problem!(model: model.name, association: association.name, problem: suggestion,
-                       associated_models: deletable_models.map(&:name))
-            elsif callback_action(association) == :skip && destroyable_models.present?
-              problem!(model: model.name, association: association.name, problem: :suggest_destroy,
-                       associated_models: destroyable_models.map(&:name))
+            case association.options[:dependent]
+            when :destroy_async
+              foreign_key = foreign_key(association.klass.table_name, model.table_name)
+              if foreign_key
```
gregnavis, line 69: It seems `foreign_key` can be inlined, however rewording the message may force use to keep that variable here. Leaving this up to you.

lib/active_record_doctor/detectors/incorrect_dependent_option.rb
```diff
@@ -30,6 +30,9 @@ def message(model:, association:, problem:, associated_models:)
 
         # rubocop:disable Layout/LineLength
         case problem
+        when :destroy_async
+          "replace `dependent: :destroy_async` on #{model}.#{association} or remove the foreign key referencing #{model} from #{association} - "\
+            "the foreign key actions will occur inside the same transaction that deletes its owner"
```
gregnavis, line 35: I suggest rewording that to:

```
don't use `dependent: :destroy_async` on #{model}.#{association} or remove the foreign key from #{table_name}.#{column_name} - associated models will be deleted in the same transaction along with #{model}
```

gregnavis (changes requested):

fatkodima:
Changed the message wording.

gregnavis (approved):

gregnavis:
@fatkodima, could you rebase it on top of `master`? Thanks!

fatkodima:
@gregnavis Rebased.

gregnavis (approved):

PR #111, opened 2022-09-15
Check column name in length validators

julianlires:
While checking each column for length validations in the model, the column name is not considered. By not doing this, the first model validation will be the one setting the conditions  for all columns in the schema.


joalbertg (approved):
Nice 🚀

PR #115, opened 2022-12-04
Require 'foreign_type' column to be non-NULL for required `belongs_to` associations

fatkodima:
For polymorphic `belongs_to` associations, both 'foreign_key' and 'foreign_type' columns should be non-NULL.


gregnavis (approved):

gregnavis:
Thank you for the contribution, @fatkodima.

PR #119, opened 2022-12-20
fix: no exists `ignore_columns` at `incorrect_dependent_option`

erick-ol:
It does not exist `ignore_columns` at `incorrect_dependent_option`, instead of it the correct one is `ignore_associations`


fatkodima (approved):

gregnavis (approved):
Great catch! Thank you for contributing.

PR #120, opened 2022-12-23
Correctly check column types when FK refers to non-PK column

bedgerotto:
Imagine you have two tables `companies` and `users` like the following
image

Let's say that, for some reason, you need to create an FK on the `entity_id` column connecting both tables
image

Considering the column in the referenced table is unique it works quite fine in terms of Database but, if you run AR Doctor on a project with these tables the result will be:

`users.entity_id references a column of a different type - foreign keys should be of the same type as the referenced column`

That's because it currently compares the FK column with the PK of the referenced table and not the column the FK refers to.

This PR intends to fix it by fetching the column based on the FK referenced column instead.

### DIFF
--- a/test/active_record_doctor/detectors/mismatched_foreign_key_type_test.rb
+++ b/test/active_record_doctor/detectors/mismatched_foreign_key_type_test.rb
@@ -15,6 +15,22 @@ class ActiveRecordDoctor::Detectors::MismatchedForeignKeyTypeTest < Minitest::Te
     OUTPUT
   end
 
+  def test_matched_foreign_key_with_non_primary_key_type_is_not_reported
+    # MySQL does not allow foreign keys to have different type than paired primary keys
+    return if mysql?
+
+    create_table(:companies, id: :bigint) do |t|
+      t.integer :entity_id
+      t.index [:entity_id], unique: true
+    end
+    create_table(:users) do |t|
+      t.references :entity, foreign_key: false, type: :integer, index: false
+      t.foreign_key :companies, column: :entity_id, primary_key: :entity_id
+    end
+
+    refute_problems
+  end
+
   def test_matched_foreign_key_type_is_not_reported
     create_table(:companies)
     create_table(:users) do |t|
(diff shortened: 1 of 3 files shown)


test/active_record_doctor/detectors/mismatched_foreign_key_type_test.rb
```diff
@@ -15,6 +15,22 @@ def test_mismatched_foreign_key_type_is_reported
     OUTPUT
   end
 
+  def test_matched_foreign_key_with_non_primary_key_type_is_not_reported
+    # MySQL does not allow foreign keys to have different type than paired primary keys
+    return if mysql?
+
+    create_table(:companies, id: :bigint) do |t|
+      t.integer :entity_id
+      t.index [:entity_id], unique: true
```
fatkodima, line 24: I think this can be simplified to just:
```ruby
t.integer :entity_id, index: true
```

There is no need for it to be `unique`.

test/active_record_doctor/detectors/mismatched_foreign_key_type_test.rb
```diff
@@ -15,6 +15,22 @@ def test_mismatched_foreign_key_type_is_reported
...
+    end
+    create_table(:users) do |t|
+      t.references :entity, foreign_key: false, type: :integer, index: false
```
fatkodima, line 27:
```ruby
t.integer :entity_id
```
?

fatkodima (approved):
Lgtm, thanks!

gregnavis:
@bedgerotto, thank you for your contribution and the detailed problem overview.

I took the liberty of applying some small changes to your PR, most importantly updating error messages to make them reflect SQL types. I'm going to merge it after the build turns out green.

bedgerotto:
@gregnavis thank you very much for the changes you added.
It seems the CI is green, do you think we can merge it? :)

gregnavis:
@bedgerotto, merged! Thank you for your contribution.

bedgerotto:
Hey @gregnavis 👋🏼
I see a few commits in the master branch since the last tag (1.10.0), do you think we could make a patch release (1.10.1) including the latest updates?
Sorry if not the proper place for asking that. Thanks in advance

gregnavis:
@bedgerotto, that's a fair point. I'm on it right now ...

gregnavis:
@bedgerotto, 1.11.0 is out! Please give it a shot and if there are any issues report them here, so that we can fix them quickly

bedgerotto:
@gregnavis thanks! I'll check it out 🙌🏼

PR #124, opened 2023-01-28
Enhance `UnindexedForeignKeys` to check for non-standard named foreign keys

fatkodima:
`UnindexedForeignKeys` currently checks only for columns named like `_id`. In most cases, these really are (or acting like without real db `FOREIGN KEY` constraint) foreign key columns. 
But this detector misses real unindexed foreign keys with non-standard names.


gregnavis:
Looks good, but needs a rebase.

gregnavis (approved):

PR #125, opened 2023-02-04
Expect unique indexes on `has_and_belongs_to_many` associations

fatkodima:
This commit enhances the missing_unique_index detector by making it aware of `has_and_belongs_to_many` associations, in addition to `has_one` associations and uniqueness validations.

This case is very often overlooked and leads to duplicates in the database or/and some unnecessary code to deal with duplicates.
I tried to imagine the case when duplicates would be useful, but was unable to do so. Instead I was able to google a couple of posts with people ranting ([example1](http://spoolz.com/2014/05/20/rails-habtm-with-unique-scope-and-select-columns/), [example2](https://www.semicolonandsons.com/code_diary/databases/has-and-belongs-to-many-tables-often-need-unique-indexes)) and suggesting to add unique indexes on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,9 +372,9 @@ Supported configuration options:
 
 ### Detecting Uniqueness Validations not Backed by an Index
 
-Model-level uniqueness validations and `has_one` associations should be backed
-by a database index in order to be robust. Otherwise you risk inserting
-duplicate values under a heavy load.
+Model-level uniqueness validations, `has_one` and `has_and_belongs_to_many`
+associations should be backed by a database index in order to be robust.
+Otherwise you risk inserting duplicate values under a heavy load.
 
 In order to detect such validations run:
 
@@ -396,6 +396,8 @@ Supported configuration options:
 - `ignore_models` - models whose uniqueness validators should not be checked.
 - `ignore_columns` - specific validators, written as Model(column1, ...), that
   should not be checked.
+- `ignore_join_tables` - join tables that should not be checked for existence
+  of unique indexes.
 
 ### Detecting Missing Non-`NULL` Constraints
 
--- a/lib/active_record_doctor/detectors/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/detectors/missing_unique_indexes.rb
@@ -13,8 +13,16 @@ module ActiveRecordDoctor
         },
         ignore_columns: {
           description: "specific validators, written as Model(column1, column2, ...), that should not be checked"
+        },
+        ignore_join_tables: {
+          description: "join tables that should not be checked for existence of unique indexes"
         }
       }
+
+      def initialize(**)
+        super
+        @reported_join_tables = []
+      end
 
       private
 
@@ -28,6 +36,8 @@ module ActiveRecordDoctor
           "without an expression index can lead to duplicates (a regular unique index is not enough)"
         when :has_ones
           "add a unique index on #{table}(#{columns.join(', ')}) - using `has_one` in #{model.name} without an index can lead to duplicates"
+        when :has_and_belongs_to_many
+          "add a unique index on #{table}(#{columns.join(', ')}) - using `has_and_belongs_to_many` in #{model.name} without an index can lead to duplicates"
         end
       end
       # rubocop:enable Layout/LineLength
@@ -35,6 +45,7 @@ module ActiveRecordDoctor
       def detect
         validations_without_indexes
         has_ones_without_indexes
+        has_and_belongs_to_many_without_indexes
       end
 
       def validations_without_indexes
@@ -106,6 +117,21 @@ module ActiveRecordDoctor
 
       def conditional_validator?(validator)
         (validator.options.keys & [:if, :unless, :conditions]).present?
+      end
+
+      def has_and_belongs_to_many_without_indexes # rubocop:disable Naming/PredicateName
+        each_model do |model|
+          each_association(model, type: :has_and_belongs_to_many, has_scope: false) do |association|
+            join_table = association.join_table
+            next if @reported_join_tables.include?(join_table) || config(:ignore_join_tables).include?(join_table)
+
+            columns = [association.foreign_key, association.association_foreign_key]
+            next if unique_index?(join_table, columns)
+
+            @reported_join_tables << join_table
+            problem!(model: model, table: join_table, columns: columns, problem: :has_and_belongs_to_many)
+          end
+        end
       end
 
       def resolve_attributes(model, attributes)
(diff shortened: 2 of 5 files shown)


lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -25,13 +25,16 @@ def message(model:, table:, columns:, problem:)
           "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
         when :has_ones
           "add a unique index on #{table}(#{columns.first}) - using `has_one` in the #{model.name} model without an index can lead to duplicates"
+        when :habtms
```
gregnavis, line 28: Let's use the fully spelled out name. I prefer to avoid abbreviations as they often introduce an extra step in understanding what is meant.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -25,13 +25,16 @@ def message(model:, table:, columns:, problem:)
...
+          "add a unique index on #{table}(#{columns.join(', ')}) - using `has_and_belongs_to_many` in the #{model.name} model without an index can lead to duplicates"
         end
       end
       # rubocop:enable Layout/LineLength
 
       def detect
         validations_without_indexes
         has_ones_without_indexes
+        habtms_without_indexes
```
gregnavis, line 37: Same as above.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -76,6 +79,21 @@ def has_ones_without_indexes # rubocop:disable Naming/PredicateName
         end
       end
 
+      def habtms_without_indexes
```
gregnavis, line 82: Same as above.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -76,6 +79,21 @@ def has_ones_without_indexes # rubocop:disable Naming/PredicateName
...
+        models.each do |model|
```
gregnavis, line 83: We should use `each_association` here.

README.md
```diff
@@ -331,7 +331,7 @@ Supported configuration options:
 
 ### Detecting Uniqueness Validations not Backed by an Index
 
-Model-level uniqueness validations and `has_one` associations should be backed
+Model-level uniqueness validations, `has_one` and `has_and_belongs_to_many` associations should be backed
```
gregnavis, line 334: Please reformat the paragraph, so that it stays below 80 characters.

gregnavis (changes requested):
@fatkodima, nice job!

It's impossible to ignore the warning emitted by that new check, and we must address this somehow, as otherwise it'd hard-break CI workflows.

I suggest adding and using the following setting:

```
ignore_join_tables: {
  description: "join tables that should not be checked for existence of unique indexes"
```

fatkodima:
Made suggested changes.

lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -100,6 +106,20 @@ def conditional_validator?(validator)
         (validator.options.keys & [:if, :unless, :conditions]).present?
       end
 
+      def has_and_belongs_to_many_without_indexes # rubocop:disable Naming/PredicateName
+        each_model do |model|
+          each_association(model, type: :has_and_belongs_to_many, has_scope: false) do |association|
+            join_table = association.join_table
+            next if config(:ignore_join_tables).include?(join_table)
+
+            columns = [association.foreign_key, association.association_foreign_key]
```
gregnavis, line 115: I think we should also check for a unique index on the two columns in reverse order as both `(a, b)` and `(b, a)` could do the job.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -389,6 +389,52 @@ def test_polymorphic_has_one_with_index
...
+      .define_model do
+        has_and_belongs_to_many :roles
+      end
+
+    create_table(:roles_users) do |t|
+      t.integer :user_id
+      t.integer :role_id
+    end
+
+    create_table(:roles).define_model
```
gregnavis, line 403: Let's also add HABTM here to be 100% sure no duplicate warnings are reported.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -389,6 +389,52 @@ def test_polymorphic_has_one_with_index
...
+    OUTPUT
+  end
+
+  def test_has_and_belongs_to_many_with_scope_and_without_index
+    create_table(:users)
+      .define_model do
+        has_and_belongs_to_many :roles, -> { unique }
+      end
+
+    create_table(:roles).define_model
```
gregnavis, line 416: Same as above.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -389,6 +389,52 @@ def test_polymorphic_has_one_with_index
...
+        has_and_belongs_to_many :roles
+      end
+
+    create_table(:roles_users) do |t|
+      t.integer :user_id
+      t.integer :role_id
+      t.index [:user_id, :role_id], unique: true
+    end
+
+    create_table(:roles).define_model
```
gregnavis, line 433: Same as above.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -440,6 +486,29 @@ def test_config_ignore_columns
...
+      .define_model do
+        has_and_belongs_to_many :roles
+      end
+
+    create_table(:roles_users) do |t|
+      t.integer :user_id
+      t.integer :role_id
+    end
+
+    create_table(:roles).define_model
```
gregnavis, line 500: Same as above.

gregnavis (changes requested):
@fatkodima, thank you for making the changes. I had another look and I _think_ there are some other improvements we should make. I missed them when reviewing the PR the first time. Also, could you rebase it?

lib/active_record_doctor/detectors/missing_unique_indexes.rb
fatkodima: This is already handled by `unique_index?` method.

gregnavis: You're right 👍🏻

gregnavis (approved):

PR #126, opened 2023-02-04
Enhance handling of polymorphic foreign key-like columns in `UnindexedForeignKeys`

fatkodima:
Accept polymorphic foreign key-like columns which are a prefix of a larger index. Previously, only exact match on index columns was made.

### DIFF
--- a/test/active_record_doctor/detectors/unindexed_foreign_keys_test.rb
+++ b/test/active_record_doctor/detectors/unindexed_foreign_keys_test.rb
@@ -10,7 +10,7 @@ class ActiveRecordDoctor::Detectors::UnindexedForeignKeysTest < Minitest::Test
     end
 
     assert_problems(<<~OUTPUT)
-      add an index on users.company_id - foreign keys are often used in database lookups and should be indexed for performance reasons
+      add an index on users(company_id) - foreign keys are often used in database lookups and should be indexed for performance reasons
     OUTPUT
   end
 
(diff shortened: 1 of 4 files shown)


test/active_record_doctor/detectors/unindexed_foreign_keys_test.rb
```diff
@@ -14,6 +14,34 @@ def test_unindexed_foreign_key_is_reported
     OUTPUT
   end
 
+  def test_unindexed_polymorphic_foreign_key_is_reported
+    skip("MySQL always indexes foreign keys") if mysql?
```
gregnavis, line 18: It seems this restriction can be removed, as we're not dealing with real foreign keys.

test/active_record_doctor/detectors/unindexed_foreign_keys_test.rb
```diff
@@ -14,6 +14,34 @@ def test_unindexed_foreign_key_is_reported
...
+
+    create_table(:notes) do |t|
+      t.integer :notable_id
+      t.string :notable_type
+    end
+
+    assert_problems(<<~OUTPUT)
+      add an index on notes.notable_id - foreign keys are often used in database lookups and should be indexed for performance reasons
```
gregnavis, line 26: This error message is incorrect and should mention `notable_type` too:

```
add an index on notes(notable_type, notable_id) - foreign keys are often used in database lookups and should be indexed for performance reasons
```

fatkodima:
Made suggested changes. The PR became a little bigger, then was initially.

fatkodima:
Rebased.

Yeah, only a few more left 😄 🎉

gregnavis (approved):

PR #133, opened 2023-04-23
Support PostgreSQL `INCLUDE` indexes in `ExtraneousIndexes` checker

fatkodima:
Rails 7.1 will have support for PostgreSQL `INCLUDE` indexes - https://github.com/rails/rails/pull/44803. If you are ok with this PR changes, we can wait when the new rails version is released or can merge earlier, since some people are running on some sort of rails head and will already benefit from this.

This gem's CI is not testing against rails head, so the newly added tests are skipped currently.


gregnavis (dismissed):
Love the idea! A rebase is needed and it's good to go!

fatkodima:
Rebased.

gregnavis (approved):

PR #135, opened 2023-06-06
Fix `extraneous_indexes` for expression indexes

fatkodima:
ActiveRecord returns `columns` for expression indexes as a string and `columns.count` raises an argument error, since is called on a string.


dogweather:
Ping

gregnavis (approved):

dogweather:
Thank you!

PR #137, opened 2023-06-07
Fix ignoring columns from `has_one` association for `missing_unique_indexes`

fatkodima:
```ruby
class User
  has_one :account
end

class Account
end
```

To disable `missing_unique_indexes` for checking unique index on `accounts.user_id`, we currently should add `ignore_columns: ["User(user_id)"]` to the config, which does not make sense. This PR fixes that so we should write as `ignore_columns: ["Account(user_id)"]`.

### DIFF
--- a/test/active_record_doctor/detectors/missing_unique_indexes_test.rb
+++ b/test/active_record_doctor/detectors/missing_unique_indexes_test.rb
@@ -601,6 +601,26 @@ class ActiveRecordDoctor::Detectors::MissingUniqueIndexesTest < Minitest::Test
     refute_problems
   end
 
+  def test_config_ignore_columns_from_has_one
+    Context.create_table(:users)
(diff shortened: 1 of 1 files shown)


gregnavis (dismissed):

gregnavis:
@fatkodima, could you rebase the PR?

fatkodima:
Rebased. Turns out, the code change is not needed now, only the test that is useful.

gregnavis:
@fatkodima, interesting! Do you happen to know why the issue is gone now?

fatkodima:
This was fixed in https://github.com/gregnavis/active_record_doctor/commit/b013fc06c917254c4679fc044721efad1477dd33 🤷 

Somehow this PR is a duplicate. We need to merge PRs faster 😅

PR #138, opened 2023-06-07
Ignore inherited validations and associations in STI models for `missing_unique_indexes`

fatkodima:
Currently, `missing_unique_indexes` reports an offense for both a parent and a child in the STI. This also requires a config to be added for both of them, if you want to ignore this offense.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_unique_indexes.rb
+++ b/lib/active_record_doctor/detectors/missing_unique_indexes.rb
@@ -22,12 +22,12 @@ module ActiveRecordDoctor
       def message(model:, table:, columns:, problem:)
         case problem
         when :validations
-          "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in the model without an index can lead to duplicates"
+          "add a unique index on #{table}(#{columns.join(', ')}) - validating uniqueness in #{model.name} without an index can lead to duplicates"
         when :case_insensitive_validations
-          "add a unique expression index on #{table}(#{columns.join(', ')}) - validating case-insensitive uniqueness in the model "\
+          "add a unique expression index on #{table}(#{columns.join(', ')}) - validating case-insensitive uniqueness in #{model.name} "\
             "without an expression index can lead to duplicates (a regular unique index is not enough)"
         when :has_ones
-          "add a unique index on #{table}(#{columns.join(', ')}) - using `has_one` in the #{model.name} model without an index can lead to duplicates"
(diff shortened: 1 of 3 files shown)


lib/active_record_doctor/detectors/missing_unique_indexes.rb
```diff
@@ -39,7 +39,9 @@ def detect
 
       def validations_without_indexes
         each_model(except: config(:ignore_models), existing_tables_only: true) do |model|
-          model.validators.each do |validator|
+          # Skip inherited validators from STI.
```
gregnavis, line 42: Let's also explain why we're skipping: Skip inherited validators from STI to prevent them from being reported multiple times on subclasses.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -55,6 +55,25 @@ def test_present_unique_index
...
+      validates :email, uniqueness: true
+    end
+
+    define_model(:Client, TransientRecord::Models::User) do
+      validates :name, uniqueness: true
+    end
+
+    assert_problems(<<~OUTPUT)
+      add a unique index on users(email) - validating uniqueness in the model without an index can lead to duplicates
+      add a unique index on users(name) - validating uniqueness in the model without an index can lead to duplicates
```
gregnavis, line 73: After taking STI into account, these error messages became confusing. It's not clear which model is defining the validator. I suggest including the model name in the message, e.g.:

```
add a unique index on users(email) - validating uniqueness in User without an index can lead to duplicates
```

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -55,6 +55,25 @@ def test_present_unique_index
     refute_problems
   end
 
+  def test_missing_unique_index_and_sti
```
gregnavis, line 58: I suggest a more descriptive test case name, e.g. `test_missing_validation_unique_index_reported_only_on_base_class`.

test/active_record_doctor/detectors/missing_unique_indexes_test.rb
```diff
@@ -344,6 +363,24 @@ def test_has_one_with_scope_and_without_index
     refute_problems
   end
 
+  def test_has_one_without_index_and_sti
```
gregnavis, line 366: I suggest a more descriptive test case name: `test_missing_has_one_unique_index_reported_only_on_base_class`.

gregnavis (changes requested):
Great contribution, @fatkodima! I left some minor comments.

gregnavis (approved):

PR #139, opened 2023-06-07
Fix `missing_non_null_constraint` when there are no STI children for STI-like parent

fatkodima:
When the STI-like parent model has no children, all the `next` branches in the `detect` method pass (specifically `next if !concrete_models.all? { |model| non_null_needed?(model, column) }`) and `MissingNonNullConstraint` erroneously adds offenses for the column. See a test case for the example.


gregnavis:
@fatkodima, I don't believe this PR is correct. I think it's perfectly valid to instantiate the base class of an STI hierarchy: in this case `type` will be `nil`. A subclass of the STI base class will have its name stored in `type`.

A proper fix for the problem would be to skip the check on the `type` column.

fatkodima:
Agreed, fixed.

gregnavis:
Perfect! Thank you, @fatkodima.

gregnavis (approved):

PR #140, opened 2023-06-09
Include table name in the message for `extraneous_indexes`

fatkodima:
I recently tested this gem on a project which has quite a few matviews in the database and a foreign wrapper for MySQL, but still is using `schema.rb` without dumped them there (don't ask why). And when `extraneous_indexes` reported extra indexes, it hasn't mentioned where these indexes are defined, so it was quite a challenge to find it through `psql`. This PR simplifies such debugging for future users.


gregnavis (approved):

gregnavis (approved):
Thanks for the PR. I took the liberty to rebase it and make a slight edit to the message.

PR #148, opened 2023-08-22
Support composite primary keys

fatkodima:
Rails 7.1+ will support composite primary keys, so I extended the gem to support this. Even though that rails 7.1 is not released yet, people on rails' `main` can already benefit from this.

I also added testing for rails `main` to the CI along the way and fixed a few incorrect gemfile specifiers in the CI config file.

### DIFF
--- a/lib/active_record_doctor/detectors/mismatched_foreign_key_type.rb
+++ b/lib/active_record_doctor/detectors/mismatched_foreign_key_type.rb
@@ -31,6 +31,7 @@ module ActiveRecordDoctor
 
             next if ignored?("#{table}.#{from_column.name}", config(:ignore_columns))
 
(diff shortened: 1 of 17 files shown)


lib/active_record_doctor/detectors/mismatched_foreign_key_type.rb
```diff
@@ -31,6 +31,7 @@ def detect
 
             next if ignored?("#{table}.#{from_column.name}", config(:ignore_columns))
 
+            # TODO: Add support for composite primary keys when https://github.com/rails/rails/pull/47637 is merged.
```
fatkodima, line 34: That PR adds a proper way to define multi-column foreign keys and getting their metadata, which we can not get correctly currently.

fatkodima: That was merged, so needs an update before merging this PR.

PR #150, opened 2023-10-08
Test against `activerecord` 7.1

fatkodima:
I made 2 changes in this PR:
1. Updated Actions config to exclude specific ruby versions (this way it will run on the latest version each time and no need to update these versions in the future); fixed some incorrect gemfiles specifications
2. Configured to test against activerecord 7.1

I would also suggest to test against rails' `HEAD`. Some people run it in production.


gregnavis:
@fatkodima, I'm going to close this PR, but have incorporated your changes in another commit. Thanks for the contribution!

PR #151, opened 2023-10-17
Fix `MissingNonNullConstraint` to work with `ActiveRecord.belongs_to_required_validates_foreign_key`

fatkodima:
Rails 7.1 introduced an optimization to required `belongs_to` association to not check for presence when foreign key's value hasn't changed (see https://github.com/rails/rails/pull/46522 for details).

In this PR I made `active_record_doctor` be aware of this configuration. We can assume, that if a validator has an `:if` condition **and** validation's message is `:required` **and** `ActiveRecord.belongs_to_required_validates_foreign_key == false`, then this condition was set with the rails itself (not a conditional set by the user).

### DIFF
--- a/lib/active_record_doctor/detectors/missing_non_null_constraint.rb
+++ b/lib/active_record_doctor/detectors/missing_non_null_constraint.rb
@@ -66,9 +66,19 @@ module ActiveRecordDoctor
         model.validators.select do |validator|
           validator.is_a?(ActiveRecord::Validations::PresenceValidator) &&
             !validator.options[:allow_nil] &&
-            !validator.options[:if] &&
-            !validator.options[:unless]
+            (required_with_if?(validator) || !conditional_validator?(validator))
         end
+      end
+
+      def required_with_if?(validator)
+        ActiveRecord.version >= Gem::Version.new("7.1") &&
+          !ActiveRecord.belongs_to_required_validates_foreign_key &&
+          validator.options[:message] == :required &&
+          validator.options[:if].present?
+      end
+
+      def conditional_validator?(validator)
+        validator.options[:if] || validator.options[:unless]
       end
     end
   end
--- a/test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
+++ b/test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
@@ -39,6 +39,28 @@ class ActiveRecordDoctor::Detectors::MissingNonNullConstraintTest < Minitest::Te
     OUTPUT
   end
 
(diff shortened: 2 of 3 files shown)


lib/active_record_doctor/detectors/missing_non_null_constraint.rb
```diff
@@ -66,10 +66,20 @@ def required_presence_validators(model)
         model.validators.select do |validator|
           validator.is_a?(ActiveRecord::Validations::PresenceValidator) &&
             !validator.options[:allow_nil] &&
-            !validator.options[:if] &&
-            !validator.options[:unless]
+            (required_with_if?(validator) || !conditional_validator?(validator))
         end
       end
+
+      def required_with_if?(validator)
```
gregnavis, line 73: If I understand correctly the intent here is checking whether `validator` was added by Rails as part of defining a `belongs_to` association. In this case, I suggest renaming this method to make the name capture that intent, e.g. `rails_belongs_to_presence_validator?`.

gregnavis (changes requested):
Thanks for the PR! It seems it adds `ignore_databases` - let's remove that, as this is a separate feature that needs to be added throughout the gem. I do have a WIP branch with multi-database support.

I left some feedback with what I believe is a better way to check whether a validator is provided by Rails.

fatkodima:
@gregnavis Applied suggested changes.

lib/active_record_doctor/detectors/missing_non_null_constraint.rb
```diff
@@ -67,14 +67,28 @@ def required_presence_validators(model)
...
+          validator.options[:message] == :required &&
+          proc_from_activerecord?(validator.options[:if])
+      end
+
+      def conditional_validator?(validator)
+        validator.options[:if] || validator.options[:unless]
+      end
+
+      def proc_from_activerecord?(object)
+        object.is_a?(Proc) && object.binding.receiver.name.start_with?("ActiveRecord")
```
gregnavis, line 90: I think we should check for `ActiveRecord::` (with two colons at the end) to avoid matching stuff added by Active Record extension gems (e.g. `ActiveRecordSearch`).

gregnavis (changes requested):
One more change and it's good to go!

lib/active_record_doctor/detectors/missing_non_null_constraint.rb
fatkodima: Changed.

gregnavis (approved):
Thank you, @fatkodima!

PR #155, opened 2023-11-11
Fix `missing_presence_validation` for custom attributes with defaults

### LINKED ISSUE #154: MissingPresenceValidation should ignore enum's with default values
If you have a model, where an enum is defined with a `default` (or `_default`_ keyword) like this:
```
class Profile < ApplicationRecord
    enum kind: {general: 0, admin: 1}, _default: :general
end

The report should not include a missing presence validator for the field.

fatkodima:
Fixes #154.

This PR fixes `missing_presence_validation` not only for `enum`s (which is internally defined as an `attribute`), but for any `attribute` having a default.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_presence_validation.rb
+++ b/lib/active_record_doctor/detectors/missing_presence_validation.rb
@@ -27,6 +27,7 @@ module ActiveRecordDoctor
           each_attribute(model, except: config(:ignore_attributes)) do |column|
             next unless validator_needed?(model, column)
             next if validator_present?(model, column)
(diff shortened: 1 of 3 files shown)


fatkodima:
Too much version crap in this PR. I would still recommend dropping support for older railses. 

`active_record_doctor` is a code quality tool. People who are on old railses do not have tests, meaning they do not care about any quality and for sure won't use this tool and are not a target audience. At least rails < 6.1 can be dropped, in my mind.

gregnavis:
I'm closing this for now as I don't believe that the original report was natural bug.

PR #157, opened 2023-11-23
Fix a false positive for `unindexed_foreign_keys` when having foreign key on a primary key column

### LINKED ISSUE #156: unindexed_foreign_keys: False positive for custom primary keys
My application uses a form of multi-table inheritance such that child table's primary key is a foreign key on the parent table.

Given the tables:

```rb
create_table "parents" do |t|
  t.string "name", null: false
end

create_table "childs", primary_key: "parent_id", id: :bigint, default: nil do |t|
  t.string "name", null: false
end

add_foreign_key "childs", "parents", column: "parent_id"
```

This will generate a table that looks like:

```
\d childs
                     Table "public.childs"
  Column   |       Type        | Collation | Nullable | Default
-----------+-------------------+-----------+----------+---------
 parent_id | bigint            |           | not null |
 name      | character varying |           | not null |
Indexes:
    "childs_pkey" PRIMARY KEY, btree (parent_id)
Foreign-key constraints:
    "fk_rails_7fda5a7c59" FOREIGN KEY (parent_id) REFERENCES parents(id)
```

As you can see, the column `parent_id` is a foreign key, has a btree index, and is also the primary key.

When I run active_record_doctor, this gets reported as missing an index:

```
$ bundle exec rails active_record_doctor:unindexed_foreign_keys
add an index on childs(parent_id) - foreign keys are often used in database lookups and should be indexed for performance reasons
```

Expected: No error.

fatkodima:
Fixes #156.

### DIFF
--- a/lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
+++ b/lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
@@ -30,6 +30,7 @@ module ActiveRecordDoctor
             next unless named_like_foreign_key?(column) || foreign_key?(table, column)
             next if indexed?(table, column)
             next if indexed_as_polymorphic?(table, column)
+            next if connection.primary_key(table) == column.name
 
             type_column_name = type_column_name(column)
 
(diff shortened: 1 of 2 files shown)


jdufresne:
@fatkodima I discovered the same false positive exists for `missing_unique_indexes`.

```
add a unique index on children(parent_id) - using `has_one` in Parent without an index can lead to duplicates
```

The primary key index is already unique and so doesn't require an additional one.

Would you I prefer I file a new issue or is this comment here sufficient?

fatkodima:
This comment is sufficient, thanks.

PR #161, opened 2023-11-24
Suggest using config file with `.rb` extension

### LINKED ISSUE #159: Idea: Allow the configuration to be named `.active_record_doctor.rb` (`.rb` suffix) to assist editors
Many editors use the filename suffix to decide how to highlight syntax and load other language hooks.

The name `.active_record_doctor` leaves this as ambiguous to an editor, so it may not automatically select Ruby syntax highlighting.

There are relatively easy workarounds to this problem by either configuring one's editor to special case `.active_record_doctor` or selecting the editor's syntax highlighting manually, but renaming the file would be more universal and friendly to new users.

Other Ruby lint tools, such as rubocop, would also pick up the suffix automatically.

fatkodima:
Fixes #159.

Backwards compatibility is preserved to also accept `.active_record_doctor` files by default (without extension), if present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ ActiveRecordDoctor::Rake::Task.new do |task|
   task.deps = []
 
   # A path to your active_record_doctor configuration file.
-  task.config_path = ::Rails.root.join(".active_record_doctor")
+  task.config_path = ::Rails.root.join(".active_record_doctor.rb")
 
   # A Proc called right before running detectors that should ensure your Active
   # Record models are preloaded and a database connection is ready.
@@ -111,7 +111,7 @@ If you want to use the default configuration then you don't have to do anything.
 Just run `active_record_doctor` in your project directory.
 
 If you want to customize the tool you should create a file named
-`.active_record_doctor` in your project root directory with content like:
+`.active_record_doctor.rb` in your project root directory with content like:
 
 ```ruby
 ActiveRecordDoctor.configure do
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ ActiveRecordDoctor::Rake::Task.new do |task|
   task.deps = []
 
   # A path to your active_record_doctor configuration file.
-  task.config_path = ::Rails.root.join(".active_record_doctor")
+  task.config_path = ::Rails.root.join(".active_record_doctor.rb")
```
jdufresne, line 58: This section it titled "Additional Installation Steps for non-Rails Projects" but this is using a Rails specific function.

gregnavis: Great catch! This needs to be fixed.

gregnavis (changes requested):
@fatkodima, I wouldn't like to support two different configuration files. Let's also show a warning when the old file is used, so that users are aware we are migrating to `.active_record_doctor.rb`.

I suggest the following wording:

```
WARNING: configuration was loaded from .active_record_doctor, but it's support will be dropped in favor of .active_record_doctor.rb. Please rename the configuration file to avoid errors in the future.
```

jdufresne:
@gregnavis Do you feel #162 solves what you're looking for? Feedback is welcome.

gregnavis:
Closing in favor of #166.

README.md
gregnavis: I opened an issue to address that. See #165.

PR #162, opened 2023-11-26
Change the default configuration file to .active_record_doctor.rb

### LINKED ISSUE #159: Idea: Allow the configuration to be named `.active_record_doctor.rb` (`.rb` suffix) to assist editors

jdufresne:
It now includes the `.rb` extension to assist editors with automatic syntax highlighting as well as other tools.

There continues to be a fallback to the previous `.active_record_doctor` for backward compatibility. If used, a deprecation warning is emitted to nudge the project towards the new standard.

Fixes #159

### DIFF
--- a/lib/active_record_doctor/rake/task.rb
+++ b/lib/active_record_doctor/rake/task.rb
@@ -69,7 +69,24 @@ module ActiveRecordDoctor
 
       def config
         @config ||= begin
-          path = config_path && File.exist?(config_path) ? config_path : nil
+          path = nil
+
+          if config_path
+            if File.exist?(config_path)
+              path = config_path
+            elsif config_path.to_s.end_with?(".rb")
+              config_path_without_extension = config_path.to_s.gsub(/\.rb\z/, "")
+              if File.exist?(config_path_without_extension)
+                warn <<~WARN.squish
+                  [DEPRECATED] The default configuration
+                  file #{config_path_without_extension} is deprecated.
+                  Use #{config_path} instead.
+                WARN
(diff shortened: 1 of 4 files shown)


lib/active_record_doctor/rake/task.rb
```diff
@@ -69,7 +69,24 @@ def runner
...
+              if File.exist?(config_path_without_extension)
+                warn <<~WARN.squish
+                  [DEPRECATED] The default configuration
+                  file #{config_path_without_extension} is deprecated.
+                  Use #{config_path} instead.
+                WARN
+                path = config_path_without_extension
+              end
+            end
+          end
```
gregnavis, line 88: This code would affect all configuration files, not only to default one. That logic should be applied only to cases where the user did not specify a configuration file and is relying on the default value instead. If the user explicitly requested an extensionless file we should not run this logic at all.

Inspired by your PR I open [my own](https://github.com/gregnavis/active_record_doctor/pull/166/files) and would like to ask you for a quick review. Thanks!

gregnavis:
Closing in favor of #166.

PR #164, opened 2023-11-28
Fix `missing_presence_validation` for polymorphic associations

### LINKED ISSUE #163: Missing Presence Validation with Polymorphic Type Column
I have a polymorphic association like:

```
belongs_to :documentable, polymorphic: true
```

There are `not_null` db constraints on the related `documentable_id` and `documentable_type` columns. When I run `active_record_doctor`, I get the following error:

```
add a `presence` validator to Document.documentable_type - it's NOT NULL but lacks a validator
```

I don't think an error should be raised since ActiveRecord already requires both the type and id columns from the single `belongs_to` association.

fatkodima:
Fixes #163.

### DIFF
--- a/test/active_record_doctor/detectors/missing_presence_validation_test.rb
+++ b/test/active_record_doctor/detectors/missing_presence_validation_test.rb
@@ -42,6 +42,65 @@ class ActiveRecordDoctor::Detectors::MissingPresenceValidationTest < Minitest::T
     refute_problems
   end
 
+  def test_non_null_column_is_not_reported_if_polymorphic_association_is_required_and_columns_are_required
+    Context.create_table(:images) do |t|
+      t.references :imageable, null: false, polymorphic: true
+    end.define_model do
+      if ActiveRecord::VERSION::MAJOR >= 6
+        belongs_to :imageable, polymorphic: true, optional: false
+      else
+        belongs_to :imageable, polymorphic: true, required: true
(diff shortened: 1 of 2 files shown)


gregnavis:
This doesn't look right. See the discussion on the corresponding issue.

test/active_record_doctor/detectors/missing_presence_validation_test.rb
```diff
@@ -42,6 +42,16 @@ def test_non_null_column_is_not_reported_if_association_validation_present
     refute_problems
   end
 
+  def test_non_null_column_is_not_reported_if_polymorphic_association_validation_present
+    Context.create_table(:images) do |t|
+      t.references :imageable, null: false, polymorphic: true
+    end.define_model do
+      belongs_to :imageable, polymorphic: true, required: true
```
gregnavis, line 49: Required is deprecated in newer versions of Rails. Let's add a conditional here and pass `required` for older versions use `optional` for newer versions.

Additionally, we should be testing for all combinations:

1. Columns are required, association is required (with covered this already)
2. Columns are required, association is optional.
3. Columns are optional, association is required.
4. Columns are optional, association is optional.

gregnavis (changes requested):

gregnavis:
I'm closing this in favor of #230 which adds support for associations.

PR #166, opened 2023-12-04
Load configuration from .active_record_doctor.rb

gregnavis:
This commit changes the default configuration file, so that text editors recognize it as a Ruby file. The old file is still supported but a warning is shown whenever it's used.


jdufresne (approved):
Thanks for following up!

gregnavis:
@fatkodima, thank you for bringing this up. This **IS** a big deal to me, as it's important for me to give credit where it's due. Let me update the PR ...

gregnavis:
@fatkodima, @jdufresne, done!

image

PR #171, opened 2024-01-28
missing_presence_validation: ignore columns with defaults

### LINKED ISSUE #170: Better handling of default values
Thank you for a lovely gem! 

I have one issue so far: how the gem suggests I add presence validation to columns with default values. I use default values because I don't care about having a presence validation, which is how I thought everyone did it.

I still want to use that rake task but managing the ignores is cumbersome. Am I doing something wrong?

fatkodima:
Fixes #170.

We can add a flag for this behaviour to the checker, but I do not think this will be needed, as this looks like a legit use case for all the cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,7 @@ Supported configuration options:
 - `ignore_models` - models whose underlying tables' columns should not be checked.
 - `ignore_attributes` - specific attributes, written as Model.attribute, that
   should not be checked.
+- `ignore_columns_with_default` - set to `true` to ignore columns with default values.
 
 ### Detecting Incorrect Presence Validations on Boolean Columns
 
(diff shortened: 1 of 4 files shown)


mhenrixon:
Cheers! Much appreciated image

mhenrixon (approved):

gregnavis:
I'd like to support @mhenrixon's use case, but at the same time I wouldn't like to force that on other users (including myself!). I suggest we add a new setting to `missing_presence_validation` such that:

1. A non-global setting `ignore_columns_with_default`.
2. The value of the new setting should be set to `false` to avoid breaking backwards compatibility.
3. If the setting is `true` then we should get the behavior suggested by @mhenrixon; if it's `false` then we should get the old behavior.

Does that make sense, @mhenrixon, @fatkodima? Let me know if I mixed something up -- I'm terribly under-slept today.

mhenrixon:
> Does that make sense, @mhenrixon, @fatkodima? Let me know if I mixed something up -- I'm terribly under-slept today.

Sounds phenomenal @gregnavis! I can relate to the undersleeping 😩

fatkodima:
Updated PR with `ignore_columns_with_default` new config option.

PR #176, opened 2024-03-18
Suggest installing the gem on `test` env too

fatkodima:
People will also use the gem on CI or in local tests, so suggest that too.


gregnavis (approved):
Great idea!

PR #180, opened 2024-03-29
missing_presence_validation: ignore counter cache columns

fatkodima:
This check produces false positives for counter cache columns. 
These are expected to be set by ActiveRecord, not by users.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_presence_validation.rb
+++ b/lib/active_record_doctor/detectors/missing_presence_validation.rb
@@ -30,6 +30,7 @@ module ActiveRecordDoctor
           each_attribute(model, except: config(:ignore_attributes)) do |column|
             next unless validator_needed?(model, column)
             next if validator_present?(model, column)
+            next if counter_cache_column?(model, column)
 
             problem!(column: column.name, model: model.name)
           end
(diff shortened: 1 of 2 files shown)


gregnavis (approved):

fatkodima:
@gregnavis Can you please merge all the open PRs?

gregnavis:
Thanks for the nudge, @fatkodima! I won't merge them all today, as I'd like to see some changes made, but I definitely need to pay attention to the outstanding PRs. I left some comments on a few other PRs.

Here, I'm not sure what to do with older Rails versions. I think we'll drop 4.2 soon (I'm sure you'd love that decision!), but that would require a new major release.

For now, I suggest the following: Rails versions that fail in CI due to outdated APIs should:

1. Show a warning to the user with the following content: "missing_presence_validation: counter cache columns will receive complaints about missing presence validations in Rails versions older than X.Y; you can add them to ignore_attributes"
2. Skip counter cache detection for those older Rails versions.

@fatkodima, does that make sense?

gregnavis:
@fatkodima, you are not lazy at all! I totally understand supporting 4.2 is a total pain in the ass.

I marked this and a few other PRs for 2.0.

PR #186, opened 2024-09-02
Check timestamp columns (created_at, updated_at) are NOT NULL

### LINKED ISSUE #184: Idea: timestamps (created_at, updated_at) should be NOT NULL
Once the timestamps columns are established on a table, they might as well be NOT NULL to enforce their existence. All ActiveRecord operations set these values so they will always exist under normal usage.

Note: for fixing old tables, a project may need to backfill data with a migration.

jdufresne:
Once the timestamps columns are established on a table, they might as well be NOT NULL to enforce their existence. All ActiveRecord operations set these values so they will always exist under normal usage.

Fixes #184

### DIFF
--- a/lib/active_record_doctor/detectors/missing_non_null_constraint.rb
+++ b/lib/active_record_doctor/detectors/missing_non_null_constraint.rb
@@ -18,8 +18,18 @@ module ActiveRecordDoctor
 
       private
 
+      TIMESTAMPS = ["created_at", "created_on", "updated_at", "updated_on"].freeze
+
       def message(column:, table:)
-        "add `NOT NULL` to #{table}.#{column} - models validates its presence but it's not non-NULL in the database"
+        if TIMESTAMPS.include?(column)
+          <<~WARN.squish
+            add `NOT NULL` to #{table}.#{column} - timestamp columns are set
+            automatically by Active Record and allowing NULL may lead to
+            inconsistencies introduced by bulk operations
+          WARN
+        else
+          "add `NOT NULL` to #{table}.#{column} - models validates its presence but it's not non-NULL in the database"
+        end
       end
 
       def detect
@@ -50,6 +60,8 @@ module ActiveRecordDoctor
       end
 
       def non_null_needed?(model, column)
+        return true if TIMESTAMPS.include?(column.name)
+
         belongs_to = model.reflect_on_all_associations(:belongs_to).find do |reflection|
           reflection.foreign_key == column.name ||
             (reflection.polymorphic? && reflection.foreign_type == column.name)
--- a/test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
+++ b/test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
@@ -310,4 +310,27 @@ class ActiveRecordDoctor::Detectors::MissingNonNullConstraintTest < Minitest::Te
 
     refute_problems
   end
+
+  def test_timestamps_with_null_are_disallowed
+    Context.create_table(:users) do |t|
+      t.timestamps null: true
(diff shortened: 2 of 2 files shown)


lib/active_record_doctor/detectors/missing_non_null_constraint.rb
```diff
@@ -19,7 +19,11 @@ class MissingNonNullConstraint < Base # :nodoc:
       private
 
       def message(column:, table:)
-        "add `NOT NULL` to #{table}.#{column} - models validates its presence but it's not non-NULL in the database"
+        message = "add `NOT NULL` to #{table}.#{column}"
+        unless ["created_at", "updated_at"].include?(column)
+          message = "#{message} - models validates its presence but it's not non-NULL in the database"
+        end
+        message
```
fatkodima, line 26: `if/else` will look easier for understanding, imo. 

Also rails considers `created_on`/`updated_on` columns the same as with `_at` suffix.

jdufresne: Thanks for the review @fatkodima. I have applied your suggestion and also added support for created_on / updated_on.

lib/active_record_doctor/detectors/missing_non_null_constraint.rb
```diff
@@ -18,8 +18,14 @@ class MissingNonNullConstraint < Base # :nodoc:
 
       private
 
+      TIMESTAMPS = ["created_at", "created_on", "updated_at", "updated_on"].freeze
+
       def message(column:, table:)
-        "add `NOT NULL` to #{table}.#{column} - models validates its presence but it's not non-NULL in the database"
+        if self.class::TIMESTAMPS.include?(column)
```
fatkodima, line 24:
```suggestion
        if TIMESTAMPS.include?(column)
```

jdufresne: @fatkodima Thanks again for the quick review! I made this change in the latest revision.

lib/active_record_doctor/detectors/missing_non_null_constraint.rb
```diff
@@ -18,8 +18,14 @@ class MissingNonNullConstraint < Base # :nodoc:
 
       private
 
+      TIMESTAMPS = ["created_at", "created_on", "updated_at", "updated_on"].freeze
+
       def message(column:, table:)
-        "add `NOT NULL` to #{table}.#{column} - models validates its presence but it's not non-NULL in the database"
+        if TIMESTAMPS.include?(column)
+          "add `NOT NULL` to #{table}.#{column}"
```
gregnavis, line 25: We should also explain to the user why we're making this recommendation. In this case, I suggest something like:

```
timestamp columns are set automatically by Active Record and allowing NULL may lead to inconsistencies introduced by bulk operations
```

gregnavis (changes requested):
Thank you for your contribution, @jdufresne. I appreciate the effort you put into creating a PR.

I left one small comment that will hopefully improve DX.

lib/active_record_doctor/detectors/missing_non_null_constraint.rb
jdufresne: Thanks for the review and suggestion. I applied this in the latest revision.

gregnavis: Thank you!

gregnavis (approved):
Thank you for the contribution, @jdufresne.

@fatkodima, could you resolve the outstanding conversations if your concerns were addressed?

fatkodima:
Lgtm 👍

PR #190, opened 2024-11-21
Fix configuration option key in README

koddsson:
The configuration is `ignore_indexes` not `ignore_columns`.

See https://github.com/gregnavis/active_record_doctor/blob/4ee897cac9d1aea7d91e28f0e3d98dcb13922405/lib/active_record_doctor/detectors/extraneous_indexes.rb#L14


gregnavis (approved):
Great catch!

PR #192, opened 2024-12-08
Require ruby 3.1+

fatkodima:
CI tests against ruby 3.1+, rails 7 requires ruby 3.1+ and rubocop.yml was also inconsistent.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,7 @@
 # We upgrade to new versions of rubocop manually so it won't harm to enable all
 # new cops.
 AllCops:
-  TargetRubyVersion: 2.5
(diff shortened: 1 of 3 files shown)


gregnavis:
@fatkodima, are you sure about Rails 7 requiring 3.1+? According [to the docs](https://guides.rubyonrails.org/v7.0/getting_started.html#installing-ruby) the minimum version is 2.7.0.

fatkodima:
Ah, yes. I looked at https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html at "Recommended Ruby Version". Still pretty old version for me. But can raise later.

gregnavis:
Got it. I'll close the PR for now, then.

PR #197, opened 2024-12-12
Add new checker for tables without default timestamp columns

fatkodima:
Just found a few tables in our app not having `created_at`/`updated_at` columns. This was done accidentally. Now, without these columns it is hard to tell if the table is used or not.

In 99.9% of cases, tables should have these columns. Without them, it is also impossible to know when the record was created/updated (which can greatly help debugging). Also the use of Rails automatic cache expiration based on updated_at column is impossible.


gregnavis (approved):
Love it!

PR #201, opened 2024-12-17
missing_non_null_constraint: ignore database views

### LINKED ISSUE #200: False positives on missing_non_null_constraint for view-backed models
I've created a virtual join table, using a postgres view via scenic. I've set up a model that uses that view, setting relationships that belongs_to another object. When I do that, `missing_non_null_constraint` tells me the IDs should have a NOT NULL constraint in the DB, but of course that's invalid.

fatkodima:
Fixes #200.

### DIFF
--- a/test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
+++ b/test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
@@ -259,6 +259,24 @@ class ActiveRecordDoctor::Detectors::MissingNonNullConstraintTest < Minitest::Te
     refute_problems
   end
 
+  def test_views_are_ignored
+    Context.create_table(:old_comments, force: true) do |t|
+      t.integer :user_id
+    end
+
+    Context.execute(<<-SQL)
+      CREATE VIEW comments AS SELECT * FROM old_comments
+    SQL
+
+    Context.define_model(:Comment) do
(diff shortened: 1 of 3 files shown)


gregnavis:
@fatkodima, I'll do some merging soon. Sorry, I've been dealing with some personal stuff.

test/active_record_doctor/detectors/missing_non_null_constraint_test.rb
```diff
@@ -259,6 +259,26 @@ def test_not_null_check_constraint
...
+      CREATE VIEW comments AS SELECT * FROM old_comments
+    SQL
+
+    Context.define_model(:Comment) do
+      belongs_to :user, required: true
+    end
+
+    refute_problems
+  ensure
+    connection.execute("DROP VIEW comments")
```
gregnavis, line 278: Same as above - we should be able to replace this with `Context.execute`.

gregnavis (changes requested):
Thanks for the PR! I left some comments on the test case.

fatkodima:
Made changes.

gregnavis:
@fatkodima, your nickname should be "lightning" 😄  Thanks for the update. Let's see whether the CI is happy with the new version of `transient_record`.

gregnavis (approved):

PR #204, opened 2025-01-02
missing_foreign_keys: consider column type for foreign key candidates

fatkodima:
The `missing_foreign_keys` checker does not currently consider the type of the column and so strings, booleans etc are marked as bad. 

We currently muted 100+ cases and 100% of these cases are false positives. I think it is reasonable to consider only integer-like and uuid columns. It is possible someone uses a string column as a foreign key, but I expect this to happen in 0.001% cases and it is better to miss this, than the need to pollute the config file with a bunch of ignores.

### DIFF
--- a/test/active_record_doctor/detectors/missing_foreign_keys_test.rb
+++ b/test/active_record_doctor/detectors/missing_foreign_keys_test.rb
@@ -21,6 +21,26 @@ class ActiveRecordDoctor::Detectors::MissingForeignKeysTest < Minitest::Test
     refute_problems
   end
 
+  def test_non_integer_missing_foreign_key_is_reported
+    Context.create_table(:users) do |t|
+      t.string :external_id
+    end
+
+    refute_problems
+  end
+
+  def test_uuid_missing_foreign_key_is_reported
+    require_uuid_column_type!
+
+    Context.create_table(:users) do |t|
+      t.uuid :company_id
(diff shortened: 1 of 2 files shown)


test/active_record_doctor/detectors/missing_foreign_keys_test.rb
```diff
@@ -21,6 +21,26 @@ def test_present_foreign_key_is_not_reported
...
+  def test_non_integer_missing_foreign_key_is_reported
+    Context.create_table(:users) do |t|
+      t.string :external_id
+    end
+
+    refute_problems
+  end
+
+  def test_uuid_missing_foreign_key_is_reported
+    skip("#{current_adapter} doesn't support materialized views") if !postgresql?
```
gregnavis, line 33: It seems this error message came from another test case. Let's use a `require_*` function from `setup.rb` here (or define one if none of the existing ones make sense).

gregnavis (changes requested):
Thanks for the PR! I left a small comment.

test/active_record_doctor/detectors/missing_foreign_keys_test.rb
fatkodima: Oops! Changed.

gregnavis (approved):

PR #206, opened 2025-01-05
unindexed_foreign_keys: consider column type for foreign key candidates

fatkodima:
Follow up to #204.

I realized, that unindexed_foreign_keys checker should do the same.

### DIFF
--- a/lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
+++ b/lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
@@ -27,7 +27,7 @@ module ActiveRecordDoctor
       def detect
         each_table(except: config(:ignore_tables)) do |table|
           each_column(table, except: config(:ignore_columns)) do |column|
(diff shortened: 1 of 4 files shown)


lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
```diff
@@ -27,7 +27,7 @@ def message(table:, columns:)
       def detect
         each_table(except: config(:ignore_tables)) do |table|
           each_column(table, except: config(:ignore_columns)) do |column|
-            next unless named_like_foreign_key?(column) || foreign_key?(table, column)
```
gregnavis, line 30: Is `named_like_foreign_key?` needed or has it become unused?

fatkodima: forgot to remove. Removed now.

gregnavis (approved):

PR #212, opened 2025-03-04
`unindexed_foreign_keys` validation for composite primary key on HABTM

benaitcheson:
### Problem
The UnindexedForeignKeys detector currently reports false positives for HABTM 
(Has And Belongs To Many) tables that use composite primary keys. It suggests 
adding indexes that already exist as part of the primary key.

### Solution
Modified the `indexed?` method to recognize columns that are part of composite 
primary keys. This prevents false positives while maintaining the original 
functionality for single-column indexes.

### Example
Given a HABTM table:
```sql
CREATE TABLE companies_users (
  company_id bigint NOT NULL,
  user_id bigint NOT NULL,
  PRIMARY KEY (company_id, user_id)
);
```

### Testing
Pointed the repo at this branch and ran the UnindexedForeignKeys command. Has fixed the issue for my repo.

### DIFF
--- a/lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
+++ b/lib/active_record_doctor/detectors/unindexed_foreign_keys.rb
@@ -54,7 +54,9 @@ module ActiveRecordDoctor
 
       def indexed?(table, column)
         connection.indexes(table).any? do |index|
-          index.columns.first == column.name
+          index.columns.first == column.name ||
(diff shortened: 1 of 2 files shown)


fatkodima:
I think this will be fixed as part of https://github.com/gregnavis/active_record_doctor/pull/148.

benaitcheson:
> I think this will be fixed as part of #148.

Amazing! Thanks @fatkodima. Love your work.

PR #213, opened 2025-03-06
incorrect_length_validation: ignore STI models

fatkodima:
If using STI and there is no validation on the model, but there is a limit in the database for the field, `incorrect_length_validation` checker produces violations for both root model and its children.

I think this is unnecessary and should be reported only for the root model, because this is the place where it should be fixed.


fatkodima:
@gregnavis Its been a while 😄. Time for another round of merges? 
We run a fork of this gem with all my PRs merged for quite a while - no problems so far.

PR #216, opened 2025-03-26
Make checkers agnostic of validator attributes types

fatkodima:
Rails validators do not convert passed attributes to either string or symbols. But some checkers assume that attributes are always of symbol type.

This is problematic for example for cases when the attribute is some interpolated string and not converted to a symbol, which I just encountered in my project.


fatkodima:
Found a bug in the implementation, need to update the PR before merging.

PR #218, opened 2025-03-26
incorrect_length_validation-check: Consider check constraints with column lenght limits

fatkodima:
In postgres, when adding a limit to some existing text column, it will exclusively lock the table (preventing reads and writes) until the column data is verified and the column is changed. This is problematic for large tables and a different approach is usually used - check constraints with length limit (e.g. `CHECK (length(email) <= 64)`).

This PR extends existing checker to account for this.

### DIFF
--- a/lib/active_record_doctor/detectors/incorrect_length_validation.rb
+++ b/lib/active_record_doctor/detectors/incorrect_length_validation.rb
@@ -33,14 +33,16 @@ module ActiveRecordDoctor
       def detect
         each_model(except: config(:ignore_models), existing_tables_only: true) do |model|
           each_attribute(model, except: config(:ignore_attributes), type: [:string, :text]) do |column|
+            table = model.table_name
             model_maximum = maximum_allowed_by_validations(model, column.name.to_sym)
-            next if model_maximum == column.limit
+            database_maximum = column_limit(table, column)
+            next if model_maximum == database_maximum
 
             problem!(
               model: model.name,
               attribute: column.name,
-              table: model.table_name,
-              database_maximum: column.limit,
+              table: table,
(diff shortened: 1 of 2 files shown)


lib/active_record_doctor/detectors/incorrect_length_validation.rb
```diff
@@ -55,6 +57,20 @@ def maximum_allowed_by_validations(model, column)
         end
         length_validator ? length_validator.options[:maximum] : nil
       end
+
+      def column_limit(table, column)
+        return column.limit if column.limit
```
gregnavis, line 62: There's an edge case we need to handle: a column can have **both** a limit and a check constraint. I suggest we add a method `check_constraint_length_limit` and pick the minimum of `column.limit` and `check_constraint_length_limit(table, column)`.

lib/active_record_doctor/detectors/incorrect_length_validation.rb
```diff
@@ -55,6 +57,20 @@ def maximum_allowed_by_validations(model, column)
...
+
+        # Example: char_length(name::text) <= 64
+        pattern = /(char_|character_)?length\(['"`]?#{column.name}(::text)?['"`]?\)\s*<=\s*(?<limit>\d+)/i
```
gregnavis, line 65: I suggest we add support for both `<` and `<=` and adjust the limit accordingly, i.e. `< 100` means the limit is `<= 99`.

gregnavis (changes requested):
I love the idea! I left a suggestion on how to improve it: support columns with both column and check limits + add support for `<`.

fatkodima:
Made changes.

PR #219, opened 2025-03-27
missing_foreign_keys: Ignore tables which records are destroyed async

fatkodima:
Currently, missing_foreign_keys erroneously suggests adding foreign keys to columns for rows, which are destroyed async from some other parent record. This PR fixes that.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_foreign_keys.rb
+++ b/lib/active_record_doctor/detectors/missing_foreign_keys.rb
@@ -31,6 +31,7 @@ module ActiveRecordDoctor
             next unless looks_like_foreign_key?(column)
             next if foreign_key?(table, column)
             next if polymorphic_foreign_key?(table, column)
+            next if model_destroyed_async?(table, column)
 
             problem!(table: table, column: column.name)
           end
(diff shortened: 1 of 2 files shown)


lib/active_record_doctor/detectors/missing_foreign_keys.rb
```diff
@@ -49,6 +50,18 @@ def polymorphic_foreign_key?(table, column)
           another_column.name == type_column_name
         end
       end
+
+      def model_destroyed_async?(table, column)
+        # Check if there are any models having `has_many ..., dependent: :destroy_async`
+        # referencing the specified model.
```
gregnavis, line 56: "referencing the specified table and column" - this method doesn't accept `model` as an argument

test/active_record_doctor/detectors/missing_foreign_keys_test.rb
```diff
@@ -41,6 +41,20 @@ def test_uuid_missing_foreign_key_is_reported
     OUTPUT
   end
 
+  def test_destroy_async_is_not_reported
+    Context.create_table(:companies).define_model do
+      class_attribute :destroy_association_async_job, default: Class.new
```
gregnavis, line 46: I understand this is normally defined when Active Job is present, but let's leave a comment stating that explicitly to avoid confusing future contributors.

gregnavis (changes requested):
@fatkodima, that's a great addition. I left some small comments.

fatkodima:
Made updates.

gregnavis (approved):

PR #220, opened 2025-03-27
incorrect_length_validation: Consider `inclusion` validations as an alternative

### LINKED ISSUE #205: Rails 7.1+ Enum Validation
Rails 7.1+ introduced validation on enums in https://github.com/rails/rails/pull/49100.

```rb
class Attachment < ApplicationRecord
  enum :attachment_type,
       { report: 'report', note: 'letter' },
       default: 'report',
       validate: true
       ...
```

When backed by a string the `incorrect_length_validation` raises:
```
the schema limits attachments.attachment_type to 255 characters but there's no length validator on Attachment.attachment_type - remove the database limit or add the validator
```

The `validates_inclusion_of` used under the hood by enum validation only accepts the options for `:allow_nil`, :allow_blank` and doesn't allow passing additional length validations.

I'd like a way to silence this without having to configure an ignore for each one. Or if possible check the enum options to ensure none of them have a length greater then the database col max when `validate: true` or `validate: { allow_nil: true }`

fatkodima:
Fixes #205.

### DIFF
--- a/lib/active_record_doctor/detectors/incorrect_length_validation.rb
+++ b/lib/active_record_doctor/detectors/incorrect_length_validation.rb
@@ -33,27 +33,46 @@ module ActiveRecordDoctor
       def detect
         each_model(except: config(:ignore_models), existing_tables_only: true) do |model|
           each_attribute(model, except: config(:ignore_attributes), type: [:string, :text]) do |column|
-            model_maximum = maximum_allowed_by_validations(model, column.name.to_sym)
-            next if model_maximum == column.limit
+            model_maximum = maximum_allowed_by_length_validation(model, column.name.to_sym)
+            database_maximum = column.limit
+            next if model_maximum == database_maximum
+            next if column.limit && covered_by_inclusion_validation?(model, column.name.to_sym, database_maximum)
 
             problem!(
               model: model.name,
               attribute: column.name,
               table: model.table_name,
-              database_maximum: column.limit,
+              database_maximum: database_maximum,
               model_maximum: model_maximum
             )
           end
         end
       end
 
-      def maximum_allowed_by_validations(model, column)
+      def maximum_allowed_by_length_validation(model, column)
         length_validator = model.validators.find do |validator|
           validator.kind == :length &&
             validator.options.include?(:maximum) &&
             validator.attributes.include?(column)
         end
         length_validator ? length_validator.options[:maximum] : nil
+      end
+
+      def covered_by_inclusion_validation?(model, column, limit)
+        inclusion_validator = model.validators.find do |validator|
+          validator.kind == :inclusion &&
+            validator.attributes.include?(column)
+        end
+
+        if inclusion_validator
+          values = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+          if values.is_a?(Array) && values.all?(String)
+            longest_value = values.max_by(&:size)
+            longest_value.size <= limit
+          end
+        else
+          false
+        end
       end
     end
   end
--- a/test/active_record_doctor/detectors/incorrect_length_validation_test.rb
+++ b/test/active_record_doctor/detectors/incorrect_length_validation_test.rb
@@ -61,6 +61,40 @@ class ActiveRecordDoctor::Detectors::IncorrectLengthValidationTest < Minitest::T
     refute_problems
   end
 
+  def test_array_inclusion_validation_with_shorter_values_and_limit_is_ok
+    Context.create_table(:users) do |t|
+      t.string :status, limit: 64
+    end.define_model do
+      validates :status, inclusion: { in: ["new", "vip"] }
+    end
+
+    refute_problems
+  end
+
+  def test_array_inclusion_validation_with_longer_values_and_limit_is_error
+    Context.create_table(:users) do |t|
+      t.string :status, limit: 64
+    end.define_model do
+      validates :status, inclusion: { in: ["new", "vip", "a" * 100] }
+    end
+
+    assert_problems(<<~OUTPUT)
+      the schema limits users.status to 64 characters but there's no length validator on Context::User.status - remove the database limit or add the validator
+    OUTPUT
+  end
+
+  def test_proc_inclusion_validation_and_limit_is_error
+    Context.create_table(:users) do |t|
+      t.string :status, limit: 64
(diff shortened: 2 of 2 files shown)


lib/active_record_doctor/detectors/incorrect_length_validation.rb
```diff
@@ -33,28 +33,47 @@ def message(model:, attribute:, table:, database_maximum:, model_maximum:)
       def detect
         each_model(except: config(:ignore_models), existing_tables_only: true) do |model|
           each_attribute(model, except: config(:ignore_attributes), type: [:string, :text]) do |column|
-            model_maximum = maximum_allowed_by_validations(model, column.name.to_sym)
-            next if model_maximum == column.limit
+            model_maximum = maximum_allowed_by_length_validation(model, column.name.to_sym)
+            database_maximum = column.limit
+            next if model_maximum == database_maximum
+            next if column.limit && covered_by_inclusion_validation?(model, column.name.to_sym, database_maximum)
```
gregnavis, line 39: Since we set `database_maximum` to `column.limit`, I suggest we don't reference `column.limit` here but use `database_maximum` consistently.

lib/active_record_doctor/detectors/incorrect_length_validation.rb
```diff
@@ -33,28 +33,47 @@ def message(model:, attribute:, table:, database_maximum:, model_maximum:)
...
         length_validator ? length_validator.options[:maximum] : nil
       end
+
+      def covered_by_inclusion_validation?(model, column, limit)
+        inclusion_validator = model.validators.find do |validator|
+          validator.kind == :inclusion &&
+            validator.attributes.include?(column)
+        end
+
+        if inclusion_validator
```
gregnavis, line 67: If it's `nil`, I suggest we return early via `return false if !inclusion_validator`, and continue with the happy case without extra indentation.

test/active_record_doctor/detectors/incorrect_length_validation_test.rb
```diff
@@ -61,6 +61,40 @@ def test_no_validation_and_no_limit_is_ok
...
+    end
+
+    refute_problems
+  end
+
+  def test_array_inclusion_validation_with_longer_values_and_limit_is_error
+    Context.create_table(:users) do |t|
+      t.string :status, limit: 64
+    end.define_model do
+      validates :status, inclusion: { in: ["new", "vip", "a" * 100] }
```
gregnavis, line 78: I suggest using the smallest length that won't work to test the boundary condition - 65 in this case.

test/active_record_doctor/detectors/incorrect_length_validation_test.rb
```diff
@@ -61,6 +61,40 @@ def test_no_validation_and_no_limit_is_ok
     refute_problems
   end
 
+  def test_array_inclusion_validation_with_shorter_values_and_limit_is_ok
+    Context.create_table(:users) do |t|
+      t.string :status, limit: 64
+    end.define_model do
+      validates :status, inclusion: { in: ["new", "vip"] }
```
gregnavis, line 68: I suggest including the longest lenght that will work - 64 in this case - to test the boundary.

lib/active_record_doctor/detectors/incorrect_length_validation.rb
```diff
@@ -33,28 +33,47 @@ def message(model:, attribute:, table:, database_maximum:, model_maximum:)
...
+          values = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+          if values.is_a?(Array) && values.all?(String)
+            longest_value = values.max_by(&:size)
+            longest_value.size <= limit
+          end
```
gregnavis, line 72: I think this can be rewritten as:

```ruby
values.is_a?(Array) && values.all? { _1.is_a?(String) && _1.size <= limit }
```

gregnavis (changes requested):
Great fix! I left some comments.

fatkodima:
Made updates.

test/active_record_doctor/detectors/incorrect_length_validation_test.rb
```diff
@@ -61,6 +61,40 @@ def test_no_validation_and_no_limit_is_ok
...
+    assert_problems(<<~OUTPUT)
+      the schema limits users.status to 64 characters but there's no length validator on Context::User.status - remove the database limit or add the validator
+    OUTPUT
+  end
+
+  def test_proc_inclusion_validation_and_limit_is_error
+    Context.create_table(:users) do |t|
+      t.string :status, limit: 64
+    end.define_model do
+      validates :status, inclusion: { in: ->(user) { STATUSES[user.country] } }
```
gregnavis, line 90: If I understand correctly we only care about the fact that `in` has received a `Proc`. If that's the case I suggest we just pass a proc with an empty body. I was a bit confused by that code in there.

lib/active_record_doctor/detectors/incorrect_length_validation.rb
```diff
@@ -33,28 +33,42 @@ def message(model:, attribute:, table:, database_maximum:, model_maximum:)
...
+      def maximum_allowed_by_length_validation(model, column)
         length_validator = model.validators.find do |validator|
           validator.kind == :length &&
             validator.options.include?(:maximum) &&
             validator.attributes.include?(column)
         end
         length_validator ? length_validator.options[:maximum] : nil
       end
+
+      def covered_by_inclusion_validation?(model, column, limit)
```
gregnavis, line 61: I think we should also handle and test the following cases:

1. `if` or `unless` passed to the validator - in this case the validator may not run in some cases, so the column should **not** be considered covered (return `false`).
2. `allow_blank` - in this case the validator would also be skipped, so `false` should be returned.

gregnavis (changes requested):
@fatkodima, I reviewed the PR again, and found two other edge cases to cover. We're almost at the finish line and I'd love to see this merged. Thank you for your work.

gregnavis:
Also, could we add a test case with an enum?

lib/active_record_doctor/detectors/incorrect_length_validation.rb
fatkodima: Added + test cases.

Issue #215, opened 2025-03-23
Suggest foreign keys only for columns with belongs_to


hatsu38:
## Summary
Currently, Active Record Doctor detects all columns ending with _id as missing foreign keys. However, not all _id columns require foreign key constraints. This leads to false positives in cases such as:

IDs used for external services (e.g., stripe_customer_id, slack_user_id)

https://github.com/gregnavis/active_record_doctor/blob/10411d43c4fe5d89f069f9388d1c959e90cf876a/lib/active_record_doctor/detectors/base.rb#L247-L252

On the other hand, detecting missing foreign keys based on belongs_to associations would provide a more accurate and meaningful result, aligning with ActiveRecord's data relationships.

## Potential Downside & Considerations
If a foreign key constraint is required but the model lacks a belongs_to association, it may not be detected.

However, in such cases, developers can resolve the issue by explicitly defining the belongs_to relationship.

Since _id-based detection results in excessive false positives, prioritizing belongs_to strikes a better balance between accuracy and usability.

Missing belongs_to associations in models can be seen as an issue beyond the scope of this gem.

fatkodima:
Similar to https://github.com/gregnavis/active_record_doctor/issues/94.

PR #221, opened 2025-03-30
[Fix #94] MissingForeignKeys detector to use belongs_to associations instead of _id suffix

### LINKED ISSUE #94: `MissingForeignKeys` detector assumptions
The checker makes an assumption that foreign-key like columns should be in the form of `..._id`. 

It can:
1. lead to false positives, when some table have some column ending with `_id` which is not a reference to other table, but, for example, some id from the external resource (`users.stripe_id`).
2. miss some custom named foreign keys: `belongs_to :user, foreign_key: :user_key`

I think, what should be really checked are **models** having non-polymorphic `belongs_to` associations without foreign keys.

Checking only models would make it easier to also ignore the case, when underlying models belong to different databases.

hatsu38:
Fixes: https://github.com/gregnavis/active_record_doctor/issues/215
Fixes: https://github.com/gregnavis/active_record_doctor/issues/94


fatkodima:
I like it, but Greg maybe not 😄

lib/active_record_doctor/detectors/missing_foreign_keys.rb
```diff
@@ -23,30 +23,30 @@ def message(table:, column:)
...
+        table_models = models.select(&:table_exists?).group_by(&:table_name)
+        views = connection.views
+
+        table_models.each do |table, models|
+          next if ignored?(table, config(:ignore_tables))
+          next if views.include?(table)
+
+          concrete_models = models.reject do |model|
+            model.abstract_class?
           end
```
gregnavis, line 35: We should be using `each_model` here, potentially enhancing it to support missing features. It's important for debugging as `each_model` prints debugging information.

lib/active_record_doctor/detectors/missing_foreign_keys.rb
```diff
@@ -23,30 +23,30 @@ def message(table:, column:)
...
 
-      def foreign_key?(table, column)
-        connection.foreign_keys(table).any? do |foreign_key|
-          foreign_key.options[:column] == column.name
-        end
-      end
+          concrete_models.each do |model|
+            foreign_keys = connection.foreign_keys(table)
+            foreign_key_columns = foreign_keys.map { |key| key.options[:column] }
+            model.reflect_on_all_associations(:belongs_to).each do |association|
```
gregnavis, line 40: There's `each_association` which we should be using here for the same reason.

gregnavis (changes requested):
Thank you for the contribution! I think this is a good idea, but the PR needs some work:

1. I left some suggestions re using `each_*` to iterate over stuff - it's important for debugging.
2. This is a breaking change, so we can't just merge it. Let's add a setting to the validator that will determine the mode in which the detector works, e.g. `scan: :column_names` would scan column names (this should be the default for backwards compatibility) and `scan: :belongs_to_associations` would run the new behavior.

You can rename `detect` to `detect_via_column_names`, add your implementation in `detect_via_belongs_to_associations` and call the right method from `detect`.

Again, thank you for the contribution. I think the mode you suggested should become the default in version 2.

lib/active_record_doctor/detectors/missing_foreign_keys.rb
hatsu38: @gregnavis Thank you! Fixed the conflicts, updated based on your feedback, and pushed again!

gregnavis (approved):

PR #222, opened 2025-04-01
missing_unique_indexes: Fix ignoring case insensitive columns

fatkodima:
Case insensitive columns in missing_unique_indexes checker were not properly ignored.


fatkodima:
I do not remember, but I think the problem was that the gem suggested adding an index on `lower(column)` and so people might think that the ignore would contain that `lower` too. `lower(column)` is not actually a column, but I think for a good ux the checker can accept both.

Issue #223, opened 2025-04-20
Global ignore is not always respected


vfonic:
I have some legacy models under `Legacy::` namespace. For example:

```ruby
class Legacy::Report < ApplicationRecord
  self.table_name = "legacy_nuovo_report"
  has_many :campioni_report, class_name: "Legacy::CampioniReport", foreign_key: "idreport", inverse_of: :report, dependent: :destroy

  validates :memo, length: { maximum: 65_535 }
end
```

Here's my `.active_record_doctor.rb` file:

```ruby
ActiveRecordDoctor.configure do
  global :ignore_tables, [/^legacy_/]
end
```

Here are some errors:

```
use `dependent: :delete_all` or similar on Legacy::Report.campioni_report - associated model Legacy::CampioniReport has no callbacks and can be deleted in bulk
the length validator on Legacy::Report.memo enforces a maximum of 65535 characters but there's no schema limit on legacy_nuovo_report.memo - remove the validator or the schema length limit
the schema limits legacy_report.campioni_qt to 255 characters but there's no length validator on Legacy::OldReport.campioni_qt - remove the database limit or add the validator
```

(the last one is a different model, but should still be ignored)

vfonic:
Oh I see, sometimes it's `ignore_tables` config options, other times it's `ignore_models`.

So here's my final `.active_record_doctor.rb` config:

```ruby
ActiveRecordDoctor.configure do
  global :ignore_tables, [/^legacy_/]
  global :ignore_models, [/^Legacy::/]
end
```

PR #224, opened 2025-04-21
Add global ignoring models example

vfonic:
As I described here: https://github.com/gregnavis/active_record_doctor/issues/223

It was confusing to me to ignore all `legacy_` tables, but still have errors about those models.
I've figured out that some checks don't use table name, but model name.

I've updated the example code to make it clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,27 +161,23 @@ obtained via the help mechanism described in the previous section.
 
 ### Regexp-Based Ignores
 
-Settings like `ignore_tables`, `ignore_indexes`, and so on accept list of
+Settings like `ignore_tables`, `ignore_indexes`, `ignore_models` and so on accept list of
 identifiers to ignore. These can be either:
 
 1. Strings - in which case an exact match is needed.
 2. Regexps - which are matched against object names, and matching ones are
    excluded from output.
 
-For example, to ignore all tables starting with `legacy_` you can write:
+For example, to ignore all tables starting with `legacy_` and all models under the `Legacy::` namespace you can write:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,28 +161,24 @@ obtained via the help mechanism described in the previous section.
...
-    "schema_migrations",
-    "active_storage_blobs",
-    "active_storage_attachments",
-    "action_text_rich_texts",
-
     # Ignore all legacy tables.
     /^legacy_/
   ]
+  global :ignore_models, [
+    /^Legacy::/
```
gregnavis, line 180: Let's add `# Ignore all legacy models.`, similarly to the case above.

gregnavis (changes requested):
Great improvement! I left one small comment. Additionally, please ensure lines are not longer than 80 characters (unless absolutely necessary).

vfonic:
Resolved.

Out of curiosity, what's the reasoning behind 80 chars line length limit?

Is this line length you use to fit the content on your (split) screen or is it just a `::Legacy` (pun intended) habit from earlier days?

I use 140 line length limit and find that to be the most comfortable _for me_.

Here's a screenshot of 80 vs 140 ruler for this project's README in VS Code (first vertical line at 80, second at 140):
Screenshot 2025-04-24 at 09 47 07

gregnavis:
> Out of curiosity, what's the reasoning behind 80 chars line length limit?

It's a bit of legacy + convenience when working with split windows. I usually have a soft limit of 80 characters and a hard limit of 120.

PR #225, opened 2025-04-28
Add Ruby 3.4 support

hatsu38:
This is a CI-only change.
Since Ruby 3.4.3 exists, we will also add CI for Ruby 3.4.


gregnavis:
Thank you for you contribution, @hatsu38. I've just let the build run to see whether it's green. If so, I'll merge your PR.

gregnavis:
It seems there are some issues with Rails 7, but I don't have time to look into this now. @hatsu38, could you have a look at what's going on there?

Issue #228, opened 2025-05-19
Suggestion: Require timestamp columns to have timezone


hlascelles:
By default, even with Rails 8, a new Rails installation will not use timestamps `with timezone` by default. They are (arguably wrongly) `without timezone`. There is some discussion of that here: https://github.com/rails/rails/pull/41084

It is agreed in that issue that it is Postgres best practice to use timezones, and Rails only uses `without timezone` defaults for legacy reasons. https://github.com/rails/rails/pull/53687

But, we can do better! If new installations won't make it the default, we can encourage it with a ARD check.

Would you accept a PR that checks to ensure timestamps are `with timezone`? It can have some helpful error info on how to make it the default:

```ruby
# config/application.rb
ActiveSupport.on_load(:active_record_postgresqladapter) do
  self.datetime_type = :timestamptz
end
```

gregnavis:
Thank you for opening the issue and providing so much context. The PRs you linked were very interesting.

Let me reiterate your proposal: we'd add a new checker that would run **only for PostgreSQL** and would report errors on all columns of the `TIMESTAMP` type. The user would be encouraged to convert them to `TIMESTAMP TO TIMEZONE`.

I could also imagine having another checker that checks the default value in the configuration and suggest a change.

Let's ensure we're on the same page before you commit to the implementation.

Again, thank you for your contribution.

fatkodima:
I need to read the linked discussions, but iirc using timestamps with timezone is a bad practice and always utc should be used and then converted to the correct timezone by the application. At least this is what I saw in all the applications I ever worked.

hlascelles:
@gregnavis yes, sounds good, converting them from `timestamp` to `timestamp with time zone`.

@fatkodima yes, don't worry this is the same thing you are saying. The values are stored in the DB as UTC and then you declare your timezone in the Rails app.

ie the perfect solution is:

```sql
CREATE TABLE foo (
  created_at timestamp with time zone NOT NULL
);
```

```ruby
config.time_zone = "Europe/London"
config.active_record.default_timezone = :utc
```

What is confusing here is that postgres has the `timestamp with time zone` type which does NOT store the timezone. Yes, that is true. All it does is [store the instant as UTC](https://stackoverflow.com/a/48069726) in the DB, then gives it back to you in a timezone based on your connection settings.

PR #229, opened 2025-07-09
Add incorrect Postgres timestamp type check

hlascelles:
By default, even with Rails 8, a new Rails installation will not use timestamps `with timezone` by default. They are (arguably wrongly) `without timezone`. There is some discussion of that here: https://github.com/rails/rails/pull/41084

It is agreed in that issue that it is Postgres best practice to use timezones, and Rails only uses `without timezone` defaults for legacy reasons. https://github.com/rails/rails/pull/53687

eg, use these settings:

```ruby
ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.datetime_type = :timestamptz

# Rails config:
config.time_zone = "Europe/London"
config.active_record.default_timezone = :utc
```

This is my first submission, I understand if they styling isn't perfect :+1: .

Also, this is the first check (?) that looks at Rails config, so I wasn't quite sure how to do that part.

Resolves: https://github.com/gregnavis/active_record_doctor/issues/228


gregnavis (changes requested):
Thank you for the submission! 🙇🏻 I left some feedback.

PR #237, opened 2026-03-18
Skip models connected to non-primary databases

### LINKED ISSUE #149: Error when listing columns for model backed by different DB
We have a couple of models that are backed by a different DB to the majority of models in our app. A rough example is something like this:
```ruby
class MainDBThings < ApplicationRecord
end
```
```ruby
class ExternalDBThings < ApplicationRecord
  establish_connection :external_db
end
```

It seems that a couple of places in `active_record_doctor` assume that all tables can be accessed from a connection to a single DB (`ActiveRecord::Base.connection`); when the "different DB" table names are accessed, a "table missing" error is raised because the connection used is to the wrong DB.

For example, this line:

https://github.com/gregnavis/active_record_doctor/blob/f3b8ba515e400b9c36e1d117ed84f91eecfd78ad/lib/active_record_doctor/detectors/missing_non_null_constraint.rb#L35

 uses `connection`, which raises an error along the lines of:
 ```
 ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "ExternalDBThings" does not exist
 ```

A possible fix might be to use `models.first.columns` instead of `connection.columns(table)` so that the per-model connection is used?

Similarly, this line has the same issue:
https://github.com/gregnavis/active_record_doctor/blob/f3b8ba515e400b9c36e1d117ed84f91eecfd78ad/lib/active_record_doctor/detectors/base.rb#L218

I'm happy to have a stab at a PR for this, but would like some guidance as to any thoughts or suggested approaches before I start. I expect the fix will be to not use `connection` without reference to a particular `model` instance.

Thanks!

tmaier:
## Summary

Fixes #149 — `active_record_doctor` crashes when the application has models
connected to a different database (e.g. a gem using `establish_connection`
to SQLite while the app uses PostgreSQL).

**Root cause:** `Base#connection` is hardcoded to `ActiveRecord::Base.connection`
(the primary database). When `missing_non_null_constraint` calls
`models.select(&:table_exists?)`, it uses each model's own connection — so a
secondary-DB model returns `true`. Then `connection.columns(table)` queries the
primary connection for that table, which doesn't exist there → crash.

**Fix:** Filter `Base#models` to exclude models whose `connection_pool` differs
from `ActiveRecord::Base.connection_pool`. This is a single guard that protects
all detectors — no per-detector changes needed.

## Changes

- **`lib/active_record_doctor/detectors/base.rb`** — Filter `models` by connection pool
- **`test/setup.rb`** — Add `SecondaryContext` for multi-DB tests; graceful adapter loading
- **`test/.../missing_non_null_constraint_test.rb`** — Test: secondary-DB model doesn't crash
- **`test/.../undefined_table_references_test.rb`** — Test: secondary-DB model isn't flagged
- **`README.md`** — Document "Multiple Databases" behavior
- **`Gemfile`** — Pin minitest ~> 5.0 (minitest 6 broke minitest-fork_executor)

## Test plan

- [x] New tests fail without the fix (RED)
- [x] New tests pass with the fix (GREEN)
- [x] Full suite: 257 runs, 0 failures, 0 errors (39 adapter-specific skips)

Disclaimer: This PR was created with the help of AI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,6 @@
 
 source "https://rubygems.org"
 gemspec
+
+# minitest 6.0 removed run_one_method which minitest-fork_executor depends on
+gem "minitest", "~> 5.0"
--- a/README.md
+++ b/README.md
@@ -184,6 +184,16 @@ ActiveRecordDoctor.configure do
 end
 ```
 
+### Multiple Databases
+
+`active_record_doctor` currently only analyzes models connected to the primary
+database. Models that use a different database connection — whether via
+`establish_connection`, `connects_to`, or by inheriting from an abstract class
+with its own connection — are automatically skipped by all detectors.
+
+This means that if you have gems or parts of your application that connect to a
+secondary database, their models won't be checked and won't cause errors.
+
 ### Indexing Unindexed Foreign Keys
 
 Foreign keys should be indexed unless it's proven ineffective. However, Rails
--- a/lib/active_record_doctor/detectors/base.rb
+++ b/lib/active_record_doctor/detectors/base.rb
@@ -156,7 +156,10 @@ module ActiveRecordDoctor
       end
 
       def models
-        ActiveRecord::Base.descendants.sort_by(&:name)
+        base_pool = ActiveRecord::Base.connection_pool
+        ActiveRecord::Base.descendants
+          .reject { |model| model.connection_pool != base_pool }
+          .sort_by(&:name)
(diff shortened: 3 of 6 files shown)


gregnavis (commented):
@tmaier, thank you for the PR.

Wouldn't adding support for multiple databases solve the problem better than skipping them?

PR #243, opened 2026-06-18
Fix missing_presence_validation crash when inclusion/exclusion :in is a method name

### LINKED ISSUE #238: missing_presence_validation crashes with NoMethodError: undefined method 'exclude?' for Symbol
> Note: This issue was written with AI assistance.

## Description

`missing_presence_validation` crashes when a model uses an inclusion or exclusion validator where `:in` is a Symbol (i.e., a method name):

```ruby
validates :status, inclusion: { in: :allowed_statuses }
```

The detector calls `.exclude?` on the Symbol value, which fails because `Symbol` does not respond to `exclude?`.

## Error

```
NoMethodError: undefined method 'exclude?' for an instance of Symbol

                  (validator_items.is_a?(Proc) || validator_items.exclude?(nil))
                                                                 ^^^^^^^^^
```

## Root cause

In `lib/active_record_doctor/detectors/missing_presence_validation.rb`, `inclusion_or_exclusion_validator_items` returns `validator.options[:in]`, which can be a Symbol when the validator delegates to a method. The calling code on lines 96 and 103 assumes the result responds to `.exclude?` / `.include?`, which Symbols do not.

## Environment

- active_record_doctor 2.0.1
- Rails 8.0

SAY-5:
Fixes #238.

`missing_presence_validation` crashes when an inclusion or exclusion validator uses a method name for `:in`, e.g. `validates :status, inclusion: { in: :allowed_statuses }`. In that case `validator.options[:in]` is a Symbol, and the detector calls `.exclude?` / `.include?` on it, raising `NoMethodError: undefined method 'exclude?' for an instance of Symbol`.

A Symbol `:in` is resolved at validation time, just like a Proc, so the allowed values aren't known statically. This treats it the same way the Proc case is already handled instead of assuming the value responds to `include?`/`exclude?`.

### DIFF
--- a/lib/active_record_doctor/detectors/missing_presence_validation.rb
+++ b/lib/active_record_doctor/detectors/missing_presence_validation.rb
@@ -93,14 +93,14 @@ module ActiveRecordDoctor
               when ActiveModel::Validations::InclusionValidator
                 validator_items = inclusion_or_exclusion_validator_items(validator)
                 (validator.attributes & attribute_names).present? &&
-                  (validator_items.is_a?(Proc) || validator_items.exclude?(nil))
+                  (dynamic_validator_items?(validator_items) || validator_items.exclude?(nil))
 
               # An exclusion validator ensures the column is not nil if it covers
               # the column and excludes nil as an allowed value explicitly.
               when ActiveModel::Validations::ExclusionValidator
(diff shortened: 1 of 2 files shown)


gregnavis (approved):
Lovely addition! Thank you for contributing.

PR #244, opened 2026-07-06
Add the latest ruby and rails to CI matrix

willnet:
Add the latest Ruby and Rails to CI, update actions/checkout to the latest version, and fix the causes of CI failures.

## CI results
- [sqlite](https://github.com/willnet/active_record_doctor/actions/runs/28781697371)
- [postgresql](https://github.com/willnet/active_record_doctor/actions/runs/28781697392)
- [mysql](https://github.com/willnet/active_record_doctor/actions/runs/28781697422)
- [lint](https://github.com/willnet/active_record_doctor/actions/runs/28781697422)


active_record_doctor.gemspec
```diff
@@ -22,12 +22,14 @@ Gem::Specification.new do |s|
 
   s.add_dependency "activerecord", ACTIVE_RECORD_SPEC
 
+  s.add_development_dependency "minitest", ">= 5.0", "< 6.0"
```
gregnavis, line 25: Why `< 6.0`? If it's because of `minitest-fork_executor` not supporting v6, I released an updated version a few days ago.

.github/workflows/mysql.yml
```diff
@@ -63,12 +63,19 @@ jobs:
           - 3306:3306
     strategy:
       matrix:
-        ruby: ["3.1", "3.2", "3.3", "3.4"]
-        active_record: ["7.0", "7.1", "7.2"]
+        ruby: ["3.1", "3.2", "3.3", "3.4", "4.0"]
+        active_record: ["7.0", "7.1", "7.2", "8.0", "8.1"]
```
gregnavis: The policy is to test only against supported versions of Rails. As of now, these are 7.2, 8.0 and 8.1. The other version can be dropped.

.github/workflows/mysql.yml
```diff
@@ -63,12 +63,19 @@ jobs:
           - 3306:3306
     strategy:
       matrix:
-        ruby: ["3.1", "3.2", "3.3", "3.4"]
-        active_record: ["7.0", "7.1", "7.2"]
+        ruby: ["3.1", "3.2", "3.3", "3.4", "4.0"]
```
gregnavis: [The currently maintained versions](https://www.ruby-lang.org/en/downloads/branches/) are 3.3, 3.4, and 4.0. The rest can be dropped.

.github/workflows/postgresql.yml
```diff
@@ -63,12 +63,19 @@ jobs:
...
-        active_record: ["7.0", "7.1", "7.2"]
+        ruby: ["3.1", "3.2", "3.3", "3.4", "4.0"]
+        active_record: ["7.0", "7.1", "7.2", "8.0", "8.1"]
+        exclude:
+          - ruby: "3.1"
+            active_record: "8.0"
+          - ruby: "3.1"
+            active_record: "8.1"
+          - ruby: "4.0"
+            active_record: "8.1"
```
gregnavis: Same comments as above.

.github/workflows/sqlite.yml
```diff
@@ -22,12 +22,19 @@ jobs:
...
+        ruby: ["3.1", "3.2", "3.3", "3.4", "4.0"]
+        active_record: ["7.0", "7.1", "7.2", "8.0", "8.1"]
+        exclude:
+          - ruby: "3.1"
+            active_record: "8.0"
+          - ruby: "3.1"
+            active_record: "8.1"
+          - ruby: "4.0"
+            active_record: "8.1"
     env:
```
gregnavis: Same comments as above.

active_record_doctor.gemspec
```diff
@@ -22,12 +22,14 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency "minitest-fork_executor", "~> 1.0.2"
   s.add_development_dependency "mysql2", "~> 0.5.6"
   s.add_development_dependency "pg", "~> 1.5.9"
   s.add_development_dependency "railties", ACTIVE_RECORD_SPEC
   s.add_development_dependency "rake", "~> 13.2.1"
+  s.add_development_dependency "rdoc"
```
gregnavis, line 31: Why is it needed?

active_record_doctor.gemspec
```diff
@@ -22,12 +22,14 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency "rubocop", "~> 1.68.0"
-  s.add_development_dependency "sqlite3", "~> 2.2.0"
+  s.add_development_dependency "sqlite3", ">= 1.4", "< 3.0"
```
gregnavis, line 33: Why `< 3.0`?

gregnavis (changes requested):
Thank you so much for your contribution. I left some comments and suggestions below.

active_record_doctor.gemspec
willnet: RDoc is no longer a default gem as of Ruby 4.0. Rails 7.0 does not depend on RDoc, so tests with Ruby 4.0 and Rails 7.0 fail at [this line in the Rakefile](https://github.com/gregnavis/active_record_doctor/blob/c42816d5467df19800c46b3cca5fbe9827848fef/Rakefile#L9)

```
/Users/willnet/ghq/github.com/gregnavis/active_record_doctor/Rakefile:9: warning: rdoc/task is found in rdoc, which is not part of the default gems since Ruby 4.0.0.
You can add rdoc to your Gemfile or gemspec to fix this error.
rake aborted!
LoadError: cannot load such file -- rdoc/task (LoadError)
/Users/willnet/ghq/github.com/gregnavis/active_record_doctor/Rakefile:9:in '<top (required)>'
/Users/willnet/.local/share/mise/installs/ruby/4.0.6/bin/bundle:25:in '<main>'
(See full trace by running task with --trace)
```

active_record_doctor.gemspec
willnet: I added `< 3.0` to prevent incompatible changes from breaking CI when sqlite3 was upgraded to version 3. If this restriction is not necessary, I can remove it.

active_record_doctor.gemspec
willnet: Even if we upgrade `minitest-fork_executor` to the latest version, v1.1.0, that alone still results in the following error. To prioritize adding support for the latest Ruby and Rails versions for now, I added a version constraint for minitest.

```
  1) Error:
ActiveRecordDoctor::AddIndexesGeneratorTest#test_create_migrations_truncates_long_index_names:
NoMethodError: undefined method 'stub' for class Time
    test/generators/active_record_doctor/add_indexes/add_indexes_generator_test.rb:112:in 'block (2 levels) in ActiveRecordDoctor::AddIndexesGeneratorTest#test_create_migrations_truncates_long_index_names'
```

If it is okay to include the minitest compatibility changes in this PR as well, I’ll add another commit.

active_record_doctor.gemspec
willnet: If CI only tests Rails 7.2 or later, the tests will pass without adding this line because Rails depends on RDoc. However, the dependency is only satisfied indirectly. Since active_record_doctor uses RDoc, I think it is better to add it explicitly.

active_record_doctor.gemspec
gregnavis: `stub` was extracted into a separate gem - `minitest-mock` - which I added to deps.

active_record_doctor.gemspec
gregnavis: We can drop `rdoc` as the documentation is really just the README file.

active_record_doctor.gemspec
gregnavis: We want CI to break if our dependencies evolve. If we don't then we won't learn a new version of `sqlite3` is incompatible with `active_record_doctor`.

gregnavis:
@willnet, thank you for your contribution. In the interest of time, I applied some changes on top of your PR and merged into `master`. I left some replies above to explain why certain decisions were made. If anything is unclear just let me know.

willnet:
@gregnavis Thank you for reviewing and merging this PR! Everything makes sense. I appreciate your feedback and adjustments.

PR #246, opened 2026-09-17
incorrect_length_validation: Recognize length: { is: N }

jom2018app:
`incorrect_length_validation` only treated `length: { maximum: N }` as a model-side length constraint. `length: { is: N }` was ignored, so a column with `limit: N` and `validates :attr, length: { is: N }` was reported as having no length validator (we have many of those so we ignore them all in `.active_record_doctor.rb` for now).

#### Details
Rails keeps `:is` as a separate option (exact length). `:in` / `:within` already work because they are rewritten to `:minimum` / `:maximum` on initialize.

This treats integer `:is` as a maximum of N so matching limits are no longer reported. A mismatched `:is` or an `:is` validator without a schema limit uses the existing messages, same as `:maximum`.


Voyagerroc-Lab (approved):
Reviewed with assistance from OpenAI Codex. Supporting `length: { is: N }` as a maximum length of `N` matches Rails validation semantics, and the added tests cover both the accepted exact bound and an overlong value.

